PR #47, opened 2023-05-10
Bot bob

alanmarkoTrilitech:
Bot bob chatbox


johnjoubert:
@alanmarkoTrilitech looking good. 

Can you make a small adjustment to account for the footer background colour:

Screenshot 2023-05-15 at 17 04 24

Perhaps a thin border or something, so that the button doesn't get lost when it's over the footer.

PR #88, opened 2023-07-11
changed last updated date to Month Year only

sashaaldrick:
* change last updated date from, for example, `5th June 2023` to `June 2023`


sashaaldrick:
I also like the current full date more tbh, even if it was a year down the line, changing 20th June 2023, to June 2023, doesn't really make it feel less outdated.

PR #107, opened 2023-09-12
Clarifying the LIGO contract tutorial

timothymcmackin:
These are my suggestions for clarifying the "Originate your first smart contract tutorial." Per discussion with Tezos folks, I'm changing the contract from using a single `main` function to separate functions for each endpoint because the `main` syntax is more complicated and soon to be deprecated.

### DIFF
--- a/src/pages/tutorials/originate-your-first-smart-contract/ligo/index.md
+++ b/src/pages/tutorials/originate-your-first-smart-contract/ligo/index.md
@@ -213,7 +213,7 @@ If it calls the reset endpoint, it does not pass any parameters.
 
    This code is an OCaml type called a *variant*, similar to an enumeration in many other languages, but with some other features.
 
-1. Add this code to create the `main` function, which defines entrypoints based on the variant in the previous step:
+1. Add this code to create the `main` function, which defines the entrypoints that the client can call:
 
    ```ocaml
    let main (action, store : parameter * storage) : operation list * storage =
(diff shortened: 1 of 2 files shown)


beatalipska:
Yo, I thought I will provide my comments here (for the original one):

- [ ] ALL links to be on `new tab`
- [ ] Prerequisites are missing octet-client
- [ ] I’m not sure if we need `Want to use SmartPy?` section
- [ ] would be good to understand `why` do need those dependencies and what is their role
- [ ] `Note Make sure you have **installed** the above CLI tools before getting started.` does not make much sense IMO
- [ ] would be good to provide a link for JsLigo and CameLigo, and OCaml (some people don’t know what it is)
- [ ] I would swap `ligo version` with `./ligo version` - the second one does not work for me 
- [ ] Run the following command to generate a local wallet with *octez-client*, making sure to replace **<my_wallet>** with a name of your choosing: — in the provided command it’s `local_wallet` not <my_wallet>
- [ ] `Use Ligo to create the contract` is not mentioning in which file we should write the main code (my assumption is `increment.mligo` but yeah… hard to say from the tutorial)
- [ ] `Originate to the Testnet` - no mention of updating the `my_tez_address` to anything, make sure to delete `<` and `>` 
- [ ] again `Calling the entrypoints` don’t mention to update the `local_wallet` value to your value
- [ ] Ok, it would be cool to have the end values of the storages provided as expected outcomes in each entrypoint: 10, 9, 0 - my assumption was that the reset was the initial value of 10, so yeah - might be useful
- [ ] Missing conclusion (but I see that you have the `Summary` part already.

timothymcmackin:
> ALL links to be on new tab

The only way to do this with vanilla markdown is to use an html `<a>` link. I will make a note to see how we could do this with a markdoc extension to flag links to open in new tabs.

> I would swap ligo version with ./ligo version - the second one does not work for me

It should work if LIGO is installed and in the user's PATH. Is that the problem?

> would be good to provide a link for JsLigo and CameLigo, and OCaml (some people don’t know what it is)

I'll work on this when I add a version for jsLIGO.

beatalipska:
> > ALL links to be on new tab
> 
> The only way to do this with vanilla markdown is to use an html `<a>` link. I will make a note to see how we could do this with a markdoc extension to flag links to open in new tabs.
Would be good to have it, not super necessary, but very useful when you're going through a tutorial that has internal links :) so thanks for taking care of it


> > I would swap ligo version with ./ligo version - the second one does not work for me
> 
> It should work if LIGO is installed and in the user's PATH. Is that the problem?
I think keeping just one method of checking the version is more than enough (the way you did that in the your text).

> > would be good to provide a link for JsLigo and CameLigo, and OCaml (some people don’t know what it is)
> 
> I'll work on this when I add a version for jsLIGO.
Great stuff, thank you!

beatalipska (dismissed):
🚀

sashaaldrick (approved):
Fantastic clarifications, thank you

beatalipska (approved):

sashaaldrick (approved):

PR #113, opened 2023-09-14
This command is different for jsligo

timothymcmackin:
The command should be:

```
ligo run dry-run increment.jsligo -m Counter -e increment "32" "10"
```


beatalipska (approved):
legit 🚀

claudebarde (approved):

PR #124, opened 2023-10-12
Archetype version of the first contract tutorial

timothymcmackin:
I tried to get this working with the mockup option in archetype, but it seems to need other CLI tools installed.


timothymcmackin:
Included upstream

PR #134, opened 2023-10-31
Clarification on tokens and types of tokens

timothymcmackin:


beatalipska (commented):
Looks very good! 🚀

timothymcmackin:
Charles reviewed offline. Merging for now, will add info on other TZIP standards and on FA2.1 when it comes out.

PR #135, opened 2023-10-31
Update smart contract tutorial to use V1 of LIGO

timothymcmackin:
- Update this tutorials to use v1 of LIGO
- Use "tez" instead of "xtz"
- I've tested with ligo 1.1.0 and it works. No need to require v 1.1.0, so it says 1.0.0 or later.

### DIFF
--- a/docs/tutorials/smart-contract/cameligo.md
+++ b/docs/tutorials/smart-contract/cameligo.md
@@ -2,7 +2,7 @@
 title: Deploy a smart contract with CameLIGO
 authors: 'John Joubert, Sasha Aldrick, Claude Barde, Tim McMackin'
 last_update:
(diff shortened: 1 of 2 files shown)


beatalipska (approved):
🚀

zamrokk:
Also ligo 1.1.0 is out

timothymcmackin:
> Also ligo 1.1.0 is out

Yes, I tested this with ligo 1.1.0 and it still works, but there's no reason to require 1.1.0.

timothymcmackin:
> If you don't need Interfaces, just 1.0.0 is fine, otherwise you will have troubles XD

Got it. I'll merge these updates now and we can keep iterating on improved versions as LIGO improves.

PR #136, opened 2023-10-31
Clarify info on connecting wallets

timothymcmackin:
Adds a topic on connecting to wallets, which is a critical task for dApp developers.

This PR also updates the use of `requestPermissions`, which no longer accepts parameters per this comment in the Taquito docs:
https://github.com/ecadlabs/taquito/blob/e34e0094539fee4c03c7f6a1120339dc67faed29/docs/wallet_API.md?plain=1#L83

> In previous versions of Beacon, you were able to set the `network` property when doing `requestPermissions()`. This behaviour was removed from Beacon, and you must now set the network when instantiating the wallet.

### DIFF
--- a/docs/dApps/wallets.md
+++ b/docs/dApps/wallets.md
@@ -11,7 +11,8 @@ The primary tools that dApps use to connect to wallets are:
 
 - Beacon: A JavaScript/TypeScript SDK for connecting to wallets, signing transactions, and sending information about this connection between connected apps
 
-  Beacon implements the [tzip-10 proposal](https://gitlab.com/tezos/tzip/-/tree/master/proposals/tzip-10), which describes an interaction standard between wallets and dApps.
+  Beacon supports many Tezos wallets seamlessly, so you don't have to write different code for each wallet that you want to support.
+  Beacon also implements the [tzip-10 proposal](https://gitlab.com/tezos/tzip/-/tree/master/proposals/tzip-10), which describes an interaction standard between wallets and dApps.
   By using this standard, a dApp that uses Beacon can send messages over a peer-to-peer communication layer to a wallet, such as allowing a user to connect with an app on one platform, such as by scanning a QR code on a mobile app, and then use the dApp with the connected wallet on another platform, such as a desktop browser.
 
   Beacon can remember the connections that have been established and the accounts that have connected to the app.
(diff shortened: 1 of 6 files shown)


docs/dApps/wallets.md
```diff
@@ -1,68 +1,125 @@
...
 last_update:
-  date: 10 July 2023
+  date: 6 November 2023
 ---
 
-## Overview of wallets
+dApps must connect to user wallets to view the tokens in the account and to submit transactions on behalf of the wallet's owner.
 
-Wallets are an essential component of using blockchain services. They allow users to see their balances, find the NFTs they own, and interact with the Tezos blockchain and its smart contracts.
+The primary tools that dApps use to connect to wallets are:
```
claudebarde, line 10: Technically speaking, Beacon is the only tool, Taquito only offers a wrapper for Beacon to make it easier to use.

docs/dApps/wallets.md
```diff
@@ -1,68 +1,125 @@
...
 
-### Desktop wallets
+- Beacon: A JavaScript/TypeScript SDK for connecting to wallets, signing transactions, and sending information about this connection between connected apps
 
-- [Temple](https://templewallet.com/) (browser extension)
-- [Kukai](https://wallet.kukai.app/) (website)
-- [Umami](https://umamiwallet.com/) (app)
+  Beacon supports many Tezos wallets seamlessly, so you don't have to write different code for each wallet that you want to support.
+  Beacon also implements the [tzip-10 proposal](https://gitlab.com/tezos/tzip/-/tree/master/proposals/tzip-10), which describes an interaction standard between wallets and dApps.
```
claudebarde, line 15: As far as I know, Beacon only supports wallets that implements the TZIP-10 standard, but it may be worth double-checking.

docs/dApps/wallets.md
timothymcmackin: "Taquito provides a wrapper for Beacon so dApps can interact with wallets and with Tezos with the same code."

docs/dApps/wallets.md
timothymcmackin: Updated to say that Beacon works with tzip-10 and WalletConnect2.0 wallets.

PR #137, opened 2023-10-31
Remove octez node stuff and simplify octez-client stuff for developers

timothymcmackin:
Docs on most parts of octez belong in the octez docs, but Tezos app developers may need to use the octez client for things on the CLI.

Preview: https://docs-staging-git-octez-client-trili-tech.vercel.app/docs/developing/octez-client

Screenshot 2023-11-01 at 3 42 42 PM


docs/developing/octez-client/transactions.md
```diff
@@ -0,0 +1,73 @@
+---
+title: Sending transactions
```
beatalipska, line 2: This section is a bit tricky as it's called `Sending transactions`, but we have way more than that here, so maybe `Blockchain interactions` or `Transaction interactions` or just `Transactions with Octez` WDYT @timothymcmackin ?

docs/developing/octez-client/accounts.md
```diff
@@ -0,0 +1,49 @@
...
+```
+
+You can see the address of the account by running this command:
+
+```bash
+octez-client show address local_account
+```
+
+The account address (technically the hash of the public key) starts with `tz1`, `tz2`, or `tz3`.
+You can use this address to send tez to this account, such as with a faucet if you are using a testnet.
```
NicNomadic, line 26:
```suggestion
You can use this address to send tez to this account, such as from a faucet if you are using a testnet.
```

docs/developing/octez-client/transactions.md
```diff
@@ -0,0 +1,73 @@
...
+authors: Tim McMackin
+lastUpdated: 27 October 2023
+---
+
+Before sending transactions with the Octez client, make sure that you are connected to an RPC node for the network that you want to use.
```
NicNomadic, line 7: it's not a permanent connection to the node, just a configuration parameter, so maybe better (just a suggestion):
```suggestion
Before sending transactions with the Octez client, make sure that your client is configured to use an RPC node for the network that you want to use.
```

NicNomadic (commented):
I only reviewed the octez-client section, which LGTM modulo a couple of remarks.

docs/developing/octez-client/transactions.md
timothymcmackin: Hmmm. I like "sending transactions" because it's a task instead of a concept and because it's simple and something that developers need to do to test contracts. Maybe "Interacting with contracts"?

beatalipska: yep, "Interacting with contracts" sounds good

docs/developing/octez-client/transactions.md
timothymcmackin: That's good, it makes this more technically accurate -- thanks!

docs/developing/octez-client/transactions.md
timothymcmackin: Changed to "Interacting with contracts"

beatalipska (approved):
🚀

PR #139, opened 2023-10-31
Topic on calling contracts from dApps

timothymcmackin:
Basic task-based information on calling contracts from dApps, covering Beacon, Taquito, and a little about octez.

Preview: https://docs-staging-git-calling-contracts-trili-tech.vercel.app/docs/dApps/sending-transactions

### DIFF
--- a/docs/dApps/sending-transactions.md
+++ b/docs/dApps/sending-transactions.md
@@ -125,30 +125,4 @@ const result = await dAppClient.requestOperation({
 ## Octez
 
 The Octez command-line client can send tez and call contracts from the command line.
-For more information about Octez, see [Getting started with Octez](../reference/octez).
-
-### Sending tez with Octez
-
-To send tez with Octez, make sure you have imported the private key for the account that you are using.
-Then, use the `transfer` command to send tez, replacing the account alias `local_wallet` with the alias for your account and `target_account` with the alias or address of the account to send tez to:
-
-```bash
-octez-client --wait none transfer 5 from local_wallet to target_account
-```
-
-:::note Sending tez
-Sending tez to an address is a special case of calling a smart contract via the `default` entrypoint.
-However, some languages have a specific syntax for sending tez to a contract that is different from the syntax to call an entrypoint.
-A call to a smart contract always includes a transfer of a certain amount of tez, even if that amount is zero.
-:::
-
-### Calling contracts with Octez
-
-To call a contract with Octez, make sure that you have imported the private key for the account that you are using.
-Then, use the `transfer` command to call the contract.
-For example, this command calls the entrypoint `increment` on a contract with the alias `my_contract` and passes the parameter `12`:
-
-```bash
-octez-client --wait none transfer 0 from local_wallet to my_contract \
-  --entrypoint "increment" --arg "12" --burn-cap 0.1
-```
+See [Interacting with contracts](../developing/octez-client/transactions).
--- a/docs/dApps/sending-transactions.md
+++ b/docs/dApps/sending-transactions.md
@@ -1,0 +1,150 @@
+---
+title: Sending transactions
+authors: "Tim McMackin"
+lastUpdated: 23 October 2023
+---
+<!-- TODO originating contracts: https://tezostaquito.io/docs/originate -->
+
+After connecting to a wallet, dApps can call smart contract entrypoints and make transactions with that wallet.
+
+These calls can include:
+
+- Sending tez to an account or smart contract
+- Calling a smart contract entrypoint
+- Originating a smart contract
+
+:::note Sending tez
+Sending tez to an address is just a special case of calling a smart contract (via the `default` entrypoint).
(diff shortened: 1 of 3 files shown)


docs/dApps/sending-transactions.md
```diff
@@ -0,0 +1,131 @@
...
+These calls can include:
+
+- Sending tez to an account or smart contract
+- Calling a smart contract entrypoint
+- Originating a smart contract
+
+:::note Sending tez
+Sending tez to an address is a special case of calling a smart contract via the `default` entrypoint.
+However, some tools have a specific syntax for sending tez to a contract that is different from the syntax to call an entrypoint.
+A call to a smart contract always includes a transfer of a certain amount of tez, even if that amount is zero.
```
claudebarde, line 20: Could you explain what you mean here? I believe a transaction with 0 tez is a transaction that didn't transfer any tez.

docs/dApps/sending-transactions.md
```diff
@@ -0,0 +1,131 @@
...
+For information about calling contracts from other contracts, see [Operations](../smart-contracts/logic/operations).
+
+## Taquito
+
+You can use the Taquito SDK to send transactions from JavaScript/TypeScript applications.
+For more information about Taquito, see [Taquito](./taquito).
+
+### Sending tez
+
+To send tez with Taquito, connect to the user's wallet and use the `Tezos.wallet.transfer` method, as in this example:
```
claudebarde, line 32: I think it would be worth mentioning that you can also use the Contract API to send tez, in addition to the Wallet API.

docs/dApps/sending-transactions.md
```diff
@@ -0,0 +1,131 @@
...
+await contract.methods.doSomething('Param 1', 25)
+  .send()
+  .then((op) => {
+    console.log(`Waiting for ${op.opHash} to be confirmed...`);
+    return op.confirmation(2).then(() => op.opHash);
+  })
+  .catch((error) => console.log(`Error: ${JSON.stringify(error, null, 2)}`));
+```
+
+IDEs may warn that the method does not exist because they are not aware of the entrypoints on the contract.
```
claudebarde, line 78: I generally get an error message in VSCode when I try to call a method that doesn't exist on an object in TypeScript, VSCode at least is aware of the entrypoint methods, I don't know about the other IDEs.

timothymcmackin: I see an error in VScode when I try to call a contract entrypoint through the wallet API. How would VScode know what entrypoints are available on a contract?
Screenshot 2023-11-13 at 9 54 31 AM

docs/dApps/sending-transactions.md
timothymcmackin: This last sentence should probably be a separate note. I'll move it. I think it's clear: when you call a smart contract entrypoint you always transfer some amount of tez , even if that amount is zero.

timothymcmackin: @claudebarde I rewrote the intro section to clarify this note -- let me know what you think.

docs/dApps/sending-transactions.md
timothymcmackin: "You can also use the Taquito Contract API to send tez in a similar way."

docs/dApps/sending-transactions.md
claudebarde: You provided the wrong example, `get` is a method on a big map abstraction, not an entrypoint.
I may be wrong as I haven't used it in a while, but I can't remember VSCode complaining about a method on `Tezos.contract.methods`. Did you experience that?

timothymcmackin: No, in this case, the contract has an entrypoint named "get".

timothymcmackin: Oh, I see -- VScode is not showing errors on Tezos.contract.methods. I'll remove this note from the docs.

docs/dApps/sending-transactions.md
```diff
@@ -0,0 +1,136 @@
...
+For example, this code calls an entrypoint named "doSomething."
+It passes parameters in the order that the contract expects them:
+
+```javascript
+import { TezosToolkit } from "@taquito/taquito";
+const Tezos = new TezosToolkit(rpcUrl);
+
+Tezos.setWalletProvider(wallet);
+const contract = await Tezos.wallet.at(contractAddress);
+await contract.methods.doSomething('Param 1', 25)
```
claudebarde, line 76: If you are using `.then`, you don't need `await`. Using `await` is preferable though, in my opinion.
```
try {
    let op = await contract.methods.doSomething('Param 1', 25).send();
    console.log(`Waiting for ${op.opHash} to be confirmed...`);
    await op.confirmation(2);
} catch (error) {
    console.log(`Error: ${JSON.stringify(error, null, 2)}`)
}
```

claudebarde (changes requested):
See comment on line 76

PR #140, opened 2023-10-31
Architecture RPC info

timothymcmackin:
Still looking into what info is needed about the RPC interface.

### DIFF
--- a/docs/architecture/rpc.md
+++ b/docs/architecture/rpc.md
@@ -1,7 +1,8 @@
 ---
 title: The RPC protocol
 authors: "Tim McMackin"
-lastUpdated: 6 November 2023
+last_update:
+  date: 6 November 2023
 ---
 
 The Tezos RPC (Remote Procedure Call) protocol is a specification for a REST API that clients use to interact with Tezos nodes and nodes use to communicate with each other.
--- a/docs/architecture/rpc.md
+++ b/docs/architecture/rpc.md
@@ -1,20 +1,20 @@
 ---
 title: The RPC protocol
 authors: "Tim McMackin"
-lastUpdated: 26 October 2023
+lastUpdated: 6 November 2023
 ---
 
 The Tezos RPC (Remote Procedure Call) protocol is a specification for a REST API that clients use to interact with Tezos nodes and nodes use to communicate with each other.
+Clients use this protocol to submit transactions and get information about the state of the blockchain, such as account balances and contract storage
 Tezos nodes act as servers and accept HTTP requests from clients and other nodes via this protocol.
-
-- Clients use it to submit transactions and get information about the state of the blockchain, such as account balances and contract storage
-- Nodes use it to bake blocks and communicate with other nodes
 
 Tezos RPC uses JSON to send and receive data, but it does not adhere to the JSON-RPC specification.
 
-## RPC nodes
+## Public and private RPC nodes
 
-Not all Tezos nodes run RPC servers, and not all RPC servers accept requests from any client.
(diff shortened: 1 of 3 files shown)


docs/architecture/rpc.md
```diff
@@ -0,0 +1,26 @@
...
+title: The RPC protocol
+authors: "Tim McMackin"
+lastUpdated: 26 October 2023
+---
+
+The Tezos RPC (Remote Procedure Call) protocol is a specification for a REST API that clients use to interact with Tezos nodes and nodes use to communicate with each other.
+Tezos nodes act as servers and accept HTTP requests from clients and other nodes via this protocol.
+
+- Clients use it to submit transactions and get information about the state of the blockchain, such as account balances and contract storage
+- Nodes use it to bake blocks and communicate with other nodes
```
NicNomadic, line 11: are you sure that node use RPCs to bake blocks? 
And they communicate with other nodes using RPCs? (normally, they use the P2P protocol)

Maybe you mean to the fact that baker daemons use RPCs to bake blocks and to communicate with the node?

docs/architecture/rpc.md
```diff
@@ -0,0 +1,26 @@
...
+
+Tezos RPC uses JSON to send and receive data, but it does not adhere to the JSON-RPC specification.
+
+## RPC nodes
+
+Not all Tezos nodes run RPC servers, and not all RPC servers accept requests from any client.
```
NicNomadic, line 17: I think all nodes have an RPC server (so there is no notion of "RPC nodes"), it's just that their RPC interface is not always open to anyone. What about renaming this section in "Public and private nodes", for instance, and:
```suggestion
The RPC interface of Tezos nodes is subject to an access policy, so not all nodes accept all requests from any client.
```

timothymcmackin: > All Tezos nodes run RPC servers, but the RPC interface is subject to an access policy. By default, RPC servers are private and do not accept all requests from every client.

docs/architecture/rpc.md
```diff
@@ -0,0 +1,26 @@
+---
+title: The RPC protocol
+authors: "Tim McMackin"
+lastUpdated: 6 November 2023
+---
+
+The Tezos RPC (Remote Procedure Call) protocol is a specification for a REST API that clients use to interact with Tezos nodes and nodes use to communicate with each other.
+Clients use this protocol to submit transactions and get information about the state of the blockchain, such as account balances and contract storage
```
NicNomadic, line 8:
```suggestion
Clients use this protocol to submit transactions and get information about the state of the blockchain, such as account balances and contract storage.
```

NicNomadic (dismissed):
LGTM

docs/architecture/rpc.md
```diff
@@ -0,0 +1,27 @@
...
+authors: "Tim McMackin"
+last_update:
+  date: 6 November 2023
+---
+
+The Tezos RPC (Remote Procedure Call) protocol is a specification for a REST API that clients use to interact with Tezos nodes and nodes use to communicate with each other.
+Clients use this protocol to submit transactions and get information about the state of the blockchain, such as account balances and contract storage.
+Tezos nodes act as servers and accept HTTP requests from clients and other nodes via this protocol.
+
+Tezos RPC uses JSON to send and receive data, but it does not adhere to the JSON-RPC specification.
```
beatalipska, line 12: Maybe worth adding the list of calls: https://tezos.gitlab.io/shell/rpc.html#rpc-index-shell

timothymcmackin: Good suggestion. Added that link.

PR #125, opened 2023-10-17
clarified from part 1 to part 4 - After Benjamin's version

LiYuxin44:

PR #151, opened 2023-11-02
Update NFT marketplace tutorial

timothymcmackin:
Rebase and further work on https://github.com/trilitech/tezos-developer-docs/pull/125.


docs/tutorials/build-an-nft-marketplace.md
```diff
@@ -0,0 +1,84 @@
...
+
+1. Optional: If you haven't worked with Tezos NFTs before, consider doing the tutorial [Create an NFT](./create-an-nft) first.
+
+1. Set up an account with Pinata if you don't have one already and get an API key and API secret.
+For instructions, see the [Configure IPFS storage](./create-an-nft/nft-taquito#configure-ipfs-storage) section of the tutorial [Create a contract and web app that mints NFTs](./create-an-nft/nft-taquito).
+
+1. Make sure that you have installed these tools:
+
+   - [npm](https://nodejs.org/en/download/): NPM is required to install the web application's dependencies
+   - [taqueria >= v0.40.0](https://github.com/ecadlabs/taqueria): Taqueria is a platform that manages the contracts and front ends for Tezos dApps
```
LiYuxin44, line 28: Shall we include this link for the taqueria documentation instead of the GitHub link?
https://taqueria.io/

LiYuxin44 (commented):
Excellent work!! This version is very clear. I have one small suggestion below.

docs/tutorials/build-an-nft-marketplace.md
timothymcmackin: Got it, thx!

zamrokk:
Hi @timothymcmackin , I will review this version as :
- I saw some weird changes like `Taqueria](https://taqueria.io/), version 0.48.0 or later`. Last Taqueria version is 0.43.0 
- I will update to last ligo/fa v1.0.9 library

zamrokk:
Hi @timothymcmackin , I reviewed this version as :
- I saw some weird changes like `Taqueria](https://taqueria.io/), version 0.48.0 or later`. Last Taqueria version is 0.43.0 
- I updated to last ligo/fa v1.0.9 library and changes

docs/tutorials/build-an-nft-marketplace.md
```diff
@@ -0,0 +1,84 @@
...
+## Prerequisites
+
+1. Optional: If you haven't worked with Tezos NFTs before, consider doing the tutorial [Create an NFT](./create-an-nft) first.
+
+1. Set up an account with Pinata if you don't have one already and get an API key and API secret.
+   For instructions, see the [Configure IPFS storage](./create-an-nft/nft-taquito#configure-ipfs-storage) section of the tutorial [Create a contract and web app that mints NFTs](./create-an-nft/nft-taquito).
+
+1. Make sure that you have installed these tools:
+
+   - [npm](https://nodejs.org/en/download/): NPM is required to install the web application's dependencies
```
claudebarde, line 27: It should rather be "NodeJS" or "NodeJS/NPM", as the provided link is the one to NodeJS

docs/tutorials/build-an-nft-marketplace.md
```diff
@@ -0,0 +1,84 @@
...
+   - [Taqueria](https://taqueria.io/), version 0.43.0 or later: Taqueria is a platform that manages the contracts and front ends for Tezos dApps
```
claudebarde, line 28: It doesn't really "manage" the front end dapps, it can make it easier to run tests, for example with Flextesa.

docs/tutorials/build-an-nft-marketplace/part-1.md
```diff
@@ -0,0 +1,941 @@
...
+   ```
+
+   The FA2 standard does not require a mint entrypoint, but you can add one if you want to allow the contract to create more tokens after it is originated.
+   If you don't include a mint entrypoint or a way to create tokens, you must initialize the storage with all of the token information when you originate the contract.
+   This mint entrypoint accepts a name, description, symbol, and IPFS URL to an image.
+   It also accepts an ID number for the token, which the front end will manage; you could also set up the contract to remember the ID number for the next token.
+
+   First, this code verifies that the transaction sender is one of the administrators.
+   Then it creates a token metadata object with information from the parameters and adds it to the `token_metadata` big-map in the storage.
+   Note that the `decimals` metadata field is set to 0 because the token is an NFT and therefore has only one unit.
```
claudebarde, line 230: I think it may be better phrased as something like "the decimal is set to zero because an NFT doesn't need decimals". Talking about "one unit" may bring developers to wonder why it isn't "1" then.

docs/tutorials/build-an-nft-marketplace/part-1.md
```diff
@@ -0,0 +1,941 @@
...
+       enqueueSnackbar(tibe.data_message, {
+         variant: "error",
+         autoHideDuration: 10000,
+       });
+     }
+   };
+   ```
+
+   This function accepts the data that the user puts in the form.
+   It uploads the image to IPFS via Pinata and gets the IPFS hash, which is the basis for the link to the published file.
```
claudebarde, line 693: The IPFS hash is the link to retrieve the content, the URL is the link to the gateway that will retrieve the content based on the hash.

docs/tutorials/build-an-nft-marketplace/part-3.md
```diff
@@ -1,23 +1,17 @@
 ---
-title: NFT Marketplace Part 3
+title: "Part 3: Managing tokens with quantities"
 last_update:
-  date: 11 October 2023
+  date: 8 November 2023
 ---
 
-## Introduction
+Because only one of each NFT can exist, they are not the right token type to represent wine bottles, which have a type and a quantity of bottles of that type.
```
claudebarde, line 7: There can be multiple editions of the same NFT, i.e. multiple NFTs with the same id and the same metadata. Many artists mint multiple editions of the same NFT on marketplaces.

docs/tutorials/build-an-nft-marketplace.md
timothymcmackin: Changed to "Node.JS and NPM".

docs/tutorials/build-an-nft-marketplace/part-1.md
timothymcmackin: "Note that the `decimals` metadata field is set to 0 because the token is an NFT and therefore doesn't need any decimal places in its quantity."

docs/tutorials/build-an-nft-marketplace/part-1.md
timothymcmackin: > It uploads the image to IPFS via Pinata and gets the IPFS hash, which identifies the published file and allows clients to request it later.

docs/tutorials/build-an-nft-marketplace/part-3.md
timothymcmackin: Hmmm. Are those different editions not different NFTs entirely, or can I really create 5 identical NFTs? Not sure how to clarify here.

timothymcmackin: Discussed with Claude offline and settled on this:

> In the previous sections, you created unique, non-interchangeable tokens. These are not the right kind of token to represent wine bottles, which have a type and a quantity of bottles of that type. So in this part, you change the application to use a single-asset template, which lets you create a single token ID with a quantity that you define.

timothymcmackin:
Merging with reviews from Claude, Ben, and Yuxin.

PR #153, opened 2023-11-06
Set up styles for nested ordered lists

timothymcmackin:
By default, docusaurus uses list numbering like this, with items being numbered like `1.ii.c`:

Screenshot 2023-11-06 at 9 39 59 AM

I think it's easier to read as `1.a.i` as in this example:
Screenshot 2023-11-06 at 9 37 59 AM


beatalipska (dismissed):
🚀

PR #159, opened 2023-11-09
Use the online IDE for the smartpy tutorial

timothymcmackin:
This version of the first smart contract tutorial uses the smartpy online IDE instead of the octez client. It may be a bit easier to follow if the reader isn't comfortable with the command line or can't install things locally, but the smartpy IDE interface is pretty complicated, so I'm not sure which version is better.

Preview: https://docs-staging-git-smartpy-tutorial-use-online-ide-trili-tech.vercel.app/tutorials/smart-contract/smartpy

After this is merged, remember  to port the updated descriptions and content such as "What is a smart contract" to the other tutorials in this section.


onedebos:
Reviewed. Looks simple enough to follow along using the IDE.

docs/tutorials/smart-contract/smartpy.md
```diff
@@ -2,29 +2,44 @@
...
+## What is a smart contract?
+
+A smart contract is a computer program that is stored on a blockchain and runs on a blockchain.
+Because the blockchain is spread across many computer nodes, you don't have to think about where to host the program or worry whether a computer will run it or not.
+Responsibility for running the contract is distributed across all of the nodes in the Tezos system, so when you deploy a smart contract, you can be confident that it will be available and unmodified when someone wants to run it.
+
+A smart contract has these parts:
+
+- It has persistent storage, data that the contract can read and write
+- It has one or more entrypoints, which are a kind of function that clients can call, like endpoints in an API or functions or methods in many programming languages
```
claudebarde, line 41: It is more accurate to say that a smart contract has code that can be executed instead of entrypoints. Maybe it can be added that the code is executed by calling one of the entrypoints, but some contracts don't have entrypoints (the "default" entrypoint is just a shortcut to make it easier to understand how it works, but there is technically no entrypoint)

docs/tutorials/smart-contract/smartpy.md
```diff
@@ -2,29 +2,44 @@
...
+- It has a Tezos account and can store tez
```
claudebarde, line 42: The contract is an account itself, a programmable account through its Michelson code, it doesn't have an account.

claudebarde (changes requested):
Please see comments

docs/tutorials/smart-contract/smartpy.md
timothymcmackin: I think for this "first smart contract" tutorial, describing the contract as having entrypoints is the clear way to do it. We should work out the explanation of the default entrypoint here:
https://github.com/trilitech/tezos-developer-docs/blob/staging/docs/smart-contracts/entrypoints.md?plain=1#L47

docs/tutorials/smart-contract/smartpy.md
timothymcmackin: > It has a Tezos account and can store tez (technically, the contract is itself a type of Tezos account, but you can think of it as a program with a Tezos account)

PR #161, opened 2023-11-13
Architecture overview

timothymcmackin:
Beginner-friendly overview of Tezos architecture. Preview: 
https://docs-staging-git-architecture-overview-trili-tech.vercel.app/architecture


docs/architecture.mdx
NicNomadic, line 1: I apologize for my quite radical recommendation, but although there are some nice ideas in this page, I think the whole page must be restarted over, because there are multiple critical issues with the current discourse, e.g.:
- the notions of "nodes" differentiated in "baking nodes" and "accuser nodes" is quite imprecise (e.g. there are nodes that neither bake nor accuse, just to consensus, and optionally serve RPCs) and leads to confusions (e.g. not only baking nodes maintain the context, but also the consensus-only nodes)
  * it's important for users to know that they can participate in the consensus without being bakers
  * although it seemed a good idea to introduce a "baking/non-baking node" abstraction, we should be very cautious not to conflict with the notions of node, baker, and accuser, that are commonly used today, by users and bakers alike. Finally, I think we'd better stick to the current notions, to avoid the risk of a big terminology chaos
- the list of components is debatable in my opinion: is blockchain data a component? At least, it's very different from the other, which are executables (= processes)
  * talking about several Tezos networks at the same time adds unnecessary complexity (the first picture could only concern one network, while some phrase could just link to the notion of main and test networks)
- the discussion about blockchain trees vs single chain also adds complexity and should be abstracted away using the notion of consensus between nodes
- the network is already quite complicated to understand, so perhaps the L1 network should be first presented alone, and then only including rollups in a separate picture (with perhaps a simplified cloud for the L1)
- the discussion of mempools is misplaced in the section about clients, as it's something internal to the nodes
- etc.
Therefore, instead of proposing fixes to this version, I propose to start anew from the proposal in <https://gitlab.com/nomadic-labs/tezos/-/blob/nic@recenter-octez-proto-doc/docs/global/architecture.rst>, and adapt it to fit some of the ideas in the current version of your page. In a first time, we can draft this back within the gdoc, before updating this PR.

NicNomadic: I added a concrete proposal in the gdoc.

timothymcmackin: Thank you for the review — I'll take a hard look at the structure and think about where to take this info.

docs/architecture.mdx
```diff
@@ -0,0 +1,159 @@
...
+
+<LucidDiagram width="640px" height="480px" src="https://lucid.app/documents/embedded/d778aa2a-ad0a-4324-b235-ed3b35742c58" id="Uz-K28y3R8gJ" />
+
+## The blockchain data
+
+Although people often use the word "blockchain" to mean the entire system, strictly speaking, a Tezos blockchain is a series of blocks of data, each connected to the previous block in the chain, beginning with the genesis block.
+The blockchain data is maintained by a network of Tezos nodes.
+Nodes reach consensus on the next block before adding it to the chain.
+
+There are several Tezos blockchains.
```
NicNomadic, line 26: Make the connection with the usual term "network" that you use right after:
```suggestion
There are several Tezos blockchains, commonly called Tezos networks.
```

docs/architecture.mdx
```diff
@@ -0,0 +1,159 @@
...
+
+The relationship between the shell and the protocol is like the relationship between an operating system and an application.
+The operating system stays stable while the application can update itself.
+In this way, Tezos can update how it works (its protocol) without requiring nodes to accept major changes to the software that they run (the shell).
+For example, nodes can update to a new protocol version without restarting the shell.
+
+### The Tezos self-amending protocol
+
+Unlike many other blockchains, Tezos is self-amending.
+Its nodes can update the protocol that controls the possible operations and how they are processed via an online governance process.
```
NicNomadic, line 72: If you don't like the comma before "via", let's separate in two sentences, to avoid the wrong reading "processed via an online governance process":
```suggestion
Its nodes can update the protocol that controls the possible operations and how they are processed; updates are performed via an online governance process.
```

docs/architecture.mdx
```diff
@@ -0,0 +1,159 @@
...
+These updates allow Tezos to adapt to new technologies and respond to user needs.
+For example, protocol upgrades have added new features like Smart Rollups and have reduced the amount of time between blocks.
+
+This protocol is responsible for interpreting the operations in each block.
+It also provides the logic that identifies erroneous blocks.
+
+Updates to the protocol can change this logic through a voting process.
```
NicNomadic, line 79: Now that the governance has been already announced, let's rephrase to avoid repetition:
```suggestion
The online governance managing protocol updates involves a voting process.
```

docs/architecture.mdx
```diff
@@ -0,0 +1,159 @@
...
+
+## Tezos clients and servers
+
+In addition to the functions of the protocol and shell, a Tezos node also acts as a server to respond to queries and requests from clients.
+A client can query the chain’s state and can inject blocks and operations into a node.
+
+Tezos uses this client-server architecture for these main reasons:
+
+- It improves security by separating the node, which is exposed to the internet, from the baker, which has access to the client keys.
+The node and the baker can be on different computers, which allows the node to manage communication and shield bakers from network attacks.
```
NicNomadic, line 103:
```suggestion
The node and the baker can be on different computers, which allows the node to manage communication and shields bakers from network attacks.
```

docs/architecture.mdx
```diff
@@ -0,0 +1,159 @@
...
+### The Octez client
+
+The Octez client is a command-line tool that developers can use for many Tezos-related tasks, including:
+
+- Deploying, calling, testing, and interacting with contracts
+- Deploying and interacting with Smart Rollups
+- Working with accounts
+- Getting contract balances and data
+- Calling RPC endpoints directly
+- Running sapling transactions
```
NicNomadic, line 131: FYI, in the Octez doc, we always spell Sapling with a capital, but up to you to choose your convention (be consistent, though).

docs/architecture.mdx
```diff
@@ -0,0 +1,159 @@
...
+When it finds a problem, it submits a denunciation to other nodes to refuse the new blocks and punish the offending node.
+
+### The Octez client
+
+The Octez client is a command-line tool that developers can use for many Tezos-related tasks, including:
+
+- Deploying, calling, testing, and interacting with contracts
+- Deploying and interacting with Smart Rollups
+- Working with accounts
+- Getting contract balances and data
```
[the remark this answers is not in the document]
timothymcmackin: Removed.

docs/architecture.mdx
```diff
@@ -0,0 +1,159 @@
...
+- Calling RPC endpoints directly
+- Running sapling transactions
+
```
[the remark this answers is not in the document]
timothymcmackin: Added item "Setting up baking operations for testing contracts"

docs/architecture.mdx
timothymcmackin: I see that opentezos capitalizes Sapling as well -- I'll make that update.

timothymcmackin:
@NicNomadic Thanks for your review - I've made updates. Do you think this is ready to merge?

docs/architecture.mdx
```diff
@@ -0,0 +1,159 @@
...
+The Tezos blockchain is composed of many Tezos nodes running around the world, complemented by other running daemons, such as bakers and accusers.
+These processes collaborate with the overall goal of maintaining the blockchain data in a decentralized manner.
+
+The Tezos nodes are the most important piece in this architecture because they maintain the system and the blockchain data.
+Users interact with nodes through many different clients, including command-line clients, wallets, and web applications.
+
+This diagram shows a high-level view of the Tezos system:
+
+<LucidDiagram width="640px" height="480px" src="https://lucid.app/documents/embedded/d778aa2a-ad0a-4324-b235-ed3b35742c58" id="Uz-K28y3R8gJ" />
```
onedebos, line 18: The lucid diagram here doesn't load for me.

NicNomadic:
> @NicNomadic Thanks for your review - I've made updates. Do you think this is ready to merge?

It looks ready for me, but given its future visibility, I asked Yann to assign some architect for a final review, just to check that we don't say anything false.

onedebos (approved):

claudebarde (approved):

NicNomadic:
> It looks ready for me, but given its future visibility, I asked Yann to assign some architect for a final review, just to check that we don't say anything false.

@timothymcmackin: Yann told me he has seen some issues to fix in this page, so please wait for his review before merging.

yurug:
About the diagram:
- The semantics of arrows is not consistent as the same kind of arrows is used to denote the successor relationship in the blockchain and the relationship between component "X does Y" (as in "Baker Bakes blocks").
- The diagram suggests that a block only consists of transactions and the rollup inbox. That's not true: a block contains also metadata as well as many other operations that are not transactions.
- The "reveal channel" is not a communication layer between rollup nodes as suggested by the same kind of double arrows as the ones used in P2P for instance. The "reveal channel" is an "input channel" to gather data from the outside in a secured way (typically because the kernel knows exactly what it is expected and that cannot break the deterministic definition of the rollup state).
- While layer 1 nodes are clearly separated from the layer 1 blockchain in the diagram, the layer 2 nodes seem to embed a kernel and an outbox... that may be specific to each node. That's a source of confusion. Rollups should have the same status as the blockchain: a conceptual entity that is run by nodes. For the sake of clarity and consistency, I suggest to separate clearly the rollup and the rollup node in the diagram.
- Strictly speaking, the link between two blocks of the blockchain is backward, not forward.
- Strictly speaking, the outbox is not calling contract, it contains latent contract calls that are triggered by layer 1 operations.
- This architecture diagram is very generic and could be applied to most blockchains: from that perspective, it does not convey a lot of information. Illustrating the key aspects of Tezos that make it Tezos would be more valuable to me. On-chain governance is an example of such aspects.

yurug:
@timothymcmackin Let's start by iterating on the diagram.

NicNomadic:
As most of Yann comments are related to the representation of L2 in the diagram, I wonder if the L2 should be elided in this generic Tezos diagram. I mean, to just figure an L2 "cloud", and detail the rollups diagram in another more specific page. This way, the current diagram only introduces some of the main L1 executables and the client/server RPC interface described in the text. It could also help to add some details about the shell/protocol separation, typical to Tezos, and also described in the text.

timothymcmackin:
ok, that's very helpful, thanks @yurug  and @NicNomadic. I've made these changes to the diagram:

- Changed the direction of the block arrows and added "Hash of previous block" to illustrate that connection
- Reworked blocks to a list of the contents instead of separate lists for operations and the smart rollup inbox
- Removed layer 2 for now. I had hoped to include it but it's making the diagram too complicated.

Not sure how we'd communicate on-chain governance here. We do need a basic diagram of Tezos architecture as a starting point for documentation on Tezos, even if the diagram could also describe other blockchains.

timothymcmackin:
New diagram: 
Tezos architecture overview

NicNomadic:
> Not sure how we'd communicate on-chain governance here. We do need a basic diagram of Tezos architecture as a starting point for documentation on Tezos, even if the diagram could also describe other blockchains.
I suggest you (make the node a bit bigger perhaps and) include in the node a smaller box called "Protocol vX" (same version for all nodes), and then in the block boxes you change the contents so as to:
- omit the previous protocol hash, which is already implied by the arrows between blocks
- always contain Metadata at the beginning
- then contain different samples of operations each time, in different orders, to suggest that these are only examples, not a fixed structure; operations may be reordered subsets of: Transaction, Consensus operation, Voting operation, Origination, Smart rollup operation.
Some of these operations could be mentioned in the text, if they are not, in the corresponding sections (e.g. Voting operations could be mentioned in section "The Tezos self-amending protocol").

docs/architecture.mdx
```diff
@@ -0,0 +1,159 @@
...
+
+Unlike many other blockchains, Tezos is self-amending.
+Its nodes can update the protocol that controls the possible operations and how they are processed; updates are performed via an online governance process.
+These updates allow Tezos to adapt to new technologies and respond to user needs.
+For example, protocol upgrades have added new features like Smart Rollups and have reduced the amount of time between blocks.
+
+This protocol is responsible for interpreting the operations in each block.
+It also provides the logic that identifies erroneous blocks.
+
+Updates to the protocol can change this logic through a voting process.
```
NicNomadic, line 79: Here you could mention Voting operations if you included them in the picture, e.g:
```suggestion
Updates to the protocol can change this logic through a voting process, using dedicated voting operations such as protocol proposals or protocol upvotes.
```

timothymcmackin:
- I put metadata at the beginning of each block
- I like having "hash of previous block" in the block contents because it makes clear what links the blocks together
- Added different examples of operations for each block
- Updated text to describe the types of operations in the diagram

NicNomadic:
The diagram seems well aligned with the text now, except for showing that the nodes are constituted from a shell and a protocol. I agree with Yann that this is one big specificity of Tezos that we must make visible straight from this diagram.

timothymcmackin:
> The diagram seems well aligned with the text now, except for showing that the nodes are constituted from a shell and a protocol. I agree with Yann that this is one big specificity of Tezos that we must make visible straight from this diagram.

Ok, I have added the protocol and shell as part of each node. I wonder if the RPC should be inside the Node box.

Screenshot 2023-11-29 at 9 24 45 AM

NicNomadic:
The RPC is fine on the node, because there are both shell RPC and protocol RPCs.
But the shell and protocol are not on the same level, as the itemized list suggests. The shell, like its name suggests, wraps the protocol inside (this is how it can make it think that there is a linear sequence of blocks, for instance). So this would be much more visible if the Protocol was a smaller box within the Shell. It could even help to use a distinct color (or nuance of grey) for one of them.

timothymcmackin:
> The RPC is fine on the node, because there are both shell RPC and protocol RPCs. But the shell and protocol are not on the same level, as the itemized list suggests. The shell, like its name suggests, wraps the protocol inside (this is how it can make it think that there is a linear sequence of blocks, for instance). So this would be much more visible if the Protocol was a smaller box within the Shell. It could even help to use a distinct color (or nuance of grey) for one of them.

Ok, how's this?
Screenshot 2023-11-29 at 12 43 18 PM

NicNomadic:
This could work, but it would be better if you just put the Protocol in the upper side within the Shell (instead of its bottom), which better corresponds to the text, which says that the shell does low-level stuff. Basically, the Protocol is implemented on top of the shell and not the reverse. Another benefit of that is that the Protocol will be isolated from the outside, which depicts the fact that the shell shows it a simplified blockchain.

timothymcmackin:
> This could work, but it would be better if you just put the Protocol in the upper side within the Shell (instead of its bottom), which better corresponds to the text, which says that the shell does low-level stuff. Basically, the Protocol is implemented on top of the shell and not the reverse. Another benefit of that is that the Protocol will be isolated from the outside, which depicts the fact that the shell shows it a simplified blockchain.

Ok, is that what you're describing? 
Screenshot 2023-11-30 at 7 50 15 AM

NicNomadic:
Yes it seems right now, for me. @yurug are there any issues left, in the text or the diagram?

timothymcmackin:
Merging with offline approvals from Yann and Nic.

Issue #189, opened 2023-11-28
[BUG] Smart Contract Languages page links don't work


Frank-en-stein:
**Describe the bug**
The links to LIGO/SmartPy/Michelson pages from the https://docs.tezos.com/smart-contracts/languages/ page don't work due to incorrect routing.

**To Reproduce**
1. Go to https://docs.tezos.com/smart-contracts/languages/
2. Click on LIGO
3. See error as URL is set to incorrect https://docs.tezos.com/smart-contracts/languages/languages/ligo

**Expected behavior**
Loaded https://docs.tezos.com/smart-contracts/languages/ligo

PR #190, opened 2023-11-28
Fix languages page broken links

Frank-en-stein:
Fixes the issue: https://github.com/trilitech/tezos-developer-docs/issues/189

Updated the relative routing links in the Languages page.


Frank-en-stein:
I was using Edge (Chromium based) which was leading me to incorrect links such as https://docs.tezos.com/smart-contracts/languages/languages/ligo

Interestingly it's not happening anymore in Edge also. So, I'll close the PR.

PR #191, opened 2023-11-28
FA2 requires that tokens actually transfer

timothymcmackin:
This PR removes mention of soulbound tokens. You could create a token type that can't be transferred, but TZIP-12 and TZIP-7 require that the tokens actually transfer when you call the transfer entrypoint, therefore they can't support soulbound tokens, so I think the smart thing to do is to remove this mention.

### DIFF
--- a/docs/architecture/tokens.md
+++ b/docs/architecture/tokens.md
@@ -2,7 +2,7 @@
 title: Tokens
 authors: "Claude Barde, Aymeric Bethencourt, Tim McMackin"
 last_update:
-  date: 1 November 2023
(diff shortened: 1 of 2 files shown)


timothymcmackin:
@claudebarde What do you think? Should we mention soulbound tokens or not?

onedebos (approved):

timothymcmackin:
After discussion with Claude, I'm going to remove this mention of non-transferrable tokens. You could hack tokens to work like this but it goes against Tezos token standards and it is a rare use case.

PR #201, opened 2023-12-04
Clarify the default entrypoint

timothymcmackin:
This PR clarifies info about smart contract logic and the default entrypoint. I've tried to include info from this comment about the default entrypoint; hopefully I haven't oversimplified or lost anything.

Preview:
https://docs-staging-git-default-entrypoint-trili-tech.vercel.app/smart-contracts/entrypoints

https://github.com/trilitech/tezos-developer-docs/blob/0df256f97079e31a592be71999e0ac047bc4fdda/docs/smart-contracts/entrypoints.md?plain=1#L45-L73

### DIFF
--- a/docs/smart-contracts/entrypoints.md
+++ b/docs/smart-contracts/entrypoints.md
@@ -2,7 +2,7 @@
 title: Entrypoints
 authors: 'Mathias Hiron (Nomadic Labs), Sasha Aldrick (TriliTech), Tim McMackin (TriliTech)'
 last_update:
-  date: 22 December 2023
+  date: 26 December 2023
 ---
 
 The entrypoints of a contract represent the different ways that it can be called, similar to a method or function in many programming languages or an endpoint of an API.
@@ -53,6 +53,9 @@ The contract in the tutorial [Create a smart contract](../tutorials/smart-contra
 Even though your higher-level code may have separate codeblocks for each entrypoint, the compiled Michelson code uses a single codeblock with a single entrypoint, known as the default entrypoint.
 This default entrypoint uses the parameter that clients pass to decide which code to run.
 
+In most cases, developers can ignore the default entrypoint and imagine that the compiled Michelson code has multiple entrypoints like the higher-level code.
+However, in some cases, you may need to consider how the contract actually decides which code to run and how clients trigger this code.
+
 For example, when you compile the contract in the tutorial [Create a smart contract](../tutorials/smart-contract) to Michelson, its first line defines the parameter type that the contract accepts:
 
 ```
@@ -61,17 +64,17 @@ parameter (or (unit %reset) (or (int %decrement) (int %increment)))
 
 To call the `reset` entrypoint, clients technically call the default entrypoint and pass the Michelson-encoded parameter `Left Unit`.
 This parameter value means that the left value of the parameter type, which is annotated `%reset`, is set to the value `Unit`, which means no value.
-The Michelson code uses the `IF_LEFT` command to check if the left value of the parameter is defined and if so, runs the `reset` entrypoint code.
+In its logic, the compiled Michelson code uses the `IF_LEFT` command to check if the left value of the parameter is defined and if so, it runs the `reset` entrypoint code.
 
-In this way, these Octez client commands are the same; one passes `Unit` to the `reset` entrypoint and the other passes `Left Unit` to the default entrypoint:
+In this way, the following Octez client commands are equivalent; one passes `Unit` to the `reset` entrypoint and the other passes `Left Unit` to the default entrypoint:
 
 ```bash
-octez-client --wait none transfer 0 from myaccount to myContract \
+octez-client --wait none transfer 0 from myAccount to myContract \
   --entrypoint 'reset' --arg 'Unit' --burn-cap 0.1
 ```
 
 ```bash
-octez-client --wait none transfer 0 from myaccount to myContract \
+octez-client --wait none transfer 0 from myAccount to myContract \
   --arg 'Left Unit' --burn-cap 0.1
 ```
 


docs/smart-contracts/entrypoints.md
```diff
@@ -12,77 +12,55 @@ The entrypoints in a Tezos smart contract must meet these specifications:
...
 
-<!-- TODO old docs say "if this entrypoint exists" ; doesn't it always exist? -->
-<!--
-  - `default` doesn't take any parameter (or more specifically, its parameter is of type `Unit`). If this entrypoint exists, it will be executed any time the contract is called without specifying any entrypoint or parameter. This is in particular the case when a user or a contract simply sends some tez to the contract, as you do when sending tez to a user.
--->
+## The default entrypoint
 
+All contracts have a default entrypoint.
+This entrypoint runs when a caller calls the contract without specifying an entrypoint, such as when a caller sends tez to the contract.
+By default, the default entrypoint accepts the tez and does nothing.
```
rafoo, line 54: No, there is not default behavior of the default entrypoint. If you want your contract to accept tez on its default entrypoint, this is something that you need to code.

docs/smart-contracts/entrypoints.md
```diff
@@ -12,77 +12,55 @@ The entrypoints in a Tezos smart contract must meet these specifications:
...
 
-## Entrypoint implementation
+You can implement the default entrypoint to control what happens when a caller calls the contract without specifying an entrypoint and to allow callers to explicitly call the default entrypoint.
+For example, if you don't want the contract to accept tez, you can fail the transaction if the caller sends tez without specifying an entrypoint.
 
-Internally, entrypoints are implemented using a variant data type.
-The contract contains a single piece of logic that branches based on the value of that variant.
-In Michelson, the different values of the variant are annotated with the name of the corresponding entrypoint.
-<!-- TODO link to variants -->
+The sender does not pass parameters to the default entrypoint, although technically it accepts a parameter of the type `Unit`, which indicates no parameter.
```
rafoo, line 59: No, the default entrypoint can take a non-unit parameter.

timothymcmackin: Updated to: "By default, the default entrypoint accepts a parameter of the type `Unit`, which indicates no parameter. If you implement the default entrypoint, it can accept any parameter that other entrypoints can accept."

docs/smart-contracts/entrypoints.md
timothymcmackin: @rafoo Can you clarify here? Is is possible to call the default entrypoint if it is not explicitly defined in the code?

onedebos (dismissed):

docs/smart-contracts/entrypoints.md
NicNomadic: You can always call the default entry point, but you have to pass it the type of argument it expects. You may pass it tez only if it accepts this.

What is the contract you deployed?

timothymcmackin: @NicNomadic I deployed a simple version of the Counter contract to KT1R2LTg3mQoLvHtUjo2xSi7RMBUJ1sJkDiD:

```
namespace Counter {
  type storage = int;
  type returnValue = [list<operation>, storage];

  // Increment entrypoint
  @entry
  const increment = (delta : int, store : storage) : returnValue =>
    [list([]), store + delta];

  // Decrement entrypoint
  @entry
  const decrement = (delta : int, store : storage) : returnValue => {
    return [list([]), store - delta];
  }

  // Reset entrypoint
  @entry
  const reset = (_p : unit, _s : storage) : returnValue =>
    [list([]), 0];
}
```

I assume that the default entrypoint takes a parameter of type unit (because I have not implemented it to accept any different parameter), so I try to call it with commands like these:

```
octez-client --wait none transfer 5 from tim to KT1R2LTg3mQoLvHtUjo2xSi7RMBUJ1sJkDiD --burn-cap 0.1
octez-client --wait none transfer 5 from tim to KT1R2LTg3mQoLvHtUjo2xSi7RMBUJ1sJkDiD --arg 'unit' --burn-cap 0.1
```

These commands return errors like this:

```
Invalid argument passed to contract KT1R2LTg3mQoLvHtUjo2xSi7RMBUJ1sJkDiD.
At (unshown) location 0, value Unit is invalid for type or unit (or int int).
At (unshown) location 0, invalid primitive Unit, only Left or
Right can be used here.
``` 

Seems like the error is prompting me to provide the parameter for the named entrypoints, so I don't know how to call the default entrypoint.

timothymcmackin: If I call 

```
octez-client --wait none transfer 5 from tim to KT1R2LTg3mQoLvHtUjo2xSi7RMBUJ1sJkDiD --arg 'Left (Unit)' --burn-cap 0.1
``` 

it triggers the `reset` entrypoint, not the default entrypoint.

NicNomadic: The default entrypoint in the compiled Michelson contract is the one (invisible in your SmartPy program) that dispatches to the different entry points, depending on the value you pass. In your last example, it calls the reset entrypoint *because* you passed it argument `Left (Unit)`; it unwraps this argument and passes just `Unit` to reset.

So, I guess that in high-level languages such as SmartPy you may say there's no default entrypoint. And as Tezos developers usually don't program in Michelson, you may perhaps tell just this, or elude the notion altogether.

timothymcmackin: ok, so there isn't really a default entrypoint that's separate from the entrypoints I define in my high-level language. Instead, contracts have only one entrypoint and the code branches based on the parameter value. How's this?

> The Michelson code for a smart contract does not divide its code into separate codeblocks for different entrypoints like higher-level languages do. Instead, the Michelson code for a contract has a single entrypoint known as the default entrypoint. This default entrypoint runs a branch of the contract code based on the parameters that clients pass to the contract.
> 
> For example, when you compile the contract in the tutorial [Create a smart contract](http://localhost:3000/tutorials/smart-contract) to Michelson, its first line defines the parameter type that the contract accepts:
> 
> `parameter (or (unit %reset) (or (int %decrement) (int %increment)))`
>
> To call the `reset` entrypoint, clients technically call the default entrypoint and pass the Michelson-encoded parameter `LEFT Unit`. This parameter value means that the left value of the parameter type, which is annotated `%reset`, is set to the value `Unit`, which means no value. The Michelson code uses the `IF_LEFT` command to check if the left value of the parameter is defined and runs the `reset` entrypoint code.
>
> Similarly, to call the `decrement` entrypoint and pass 5, clients call the default entrypoint and pass `Right (Left 5)`. This parameter value puts the value 5 in the integer value that is annotated `%decrement`. Then the code branches based on this parameter value.

docs/smart-contracts/entrypoints.md
```diff
@@ -40,26 +40,33 @@ An entrypoint can call other entrypoints in its contract or entrypoints in other
...
 | `decrement` | `decrement` takes an `int` as a parameter and subtracts it from the previous value of the storage |
+| `reset` | `reset` takes "Unit" as a parameter and sets the storage to 0 |
 
-## The default entrypoint
+## Implementation details
 
-All contracts have a default entrypoint.
-This entrypoint runs when a caller calls the contract without specifying an entrypoint, such as when a caller sends tez to the contract.
-By default, the default entrypoint accepts the tez and does nothing.
+The Michelson code for a smart contract does not divide its code into separate codeblocks for different entrypoints like higher-level languages do.
```
NicNomadic, line 53:
```suggestion
The Michelson code compiled from a smart contract does not divide its code into separate codeblocks for different entrypoints like higher-level languages do.
```

docs/smart-contracts/entrypoints.md
```diff
@@ -40,26 +40,33 @@ An entrypoint can call other entrypoints in its contract or entrypoints in other
...
+Instead, the Michelson code for a contract has a single entrypoint known as the default entrypoint.
```
NicNomadic, line 54:
```suggestion
Instead, the compiled Michelson code for a contract has a single entrypoint known as the default entrypoint.
```

NicNomadic (changes requested):

docs/smart-contracts/entrypoints.md
NicNomadic: Yes, I assume this is the kind of code produced by any compiler to Michelson (but I didn't check this).

Note however that when you directly write Michelson code, you may deviate from this in at least two ways:
1. you are not forced to use IF_LEFT/IF_RIGHT to check your parameter and branch to a corresponding code section; for example, you can ignore the parameter and always call the same code, without any value; or you can use IF_LEFT/IF_RIGHT to compute some integer value out of it, and always call the same code with this value
2. you may annotate some branch of the parameter with the name "default": in this case, the root of the OR type may never be used, unless it is named differently!
But okay, these cases are purely theoretical, and pretty contorted, so it is out of question to describe them in the devs doc. Even, I wonder if all the "Implementation details" section makes sense for them. Or you may leave it but warn people that these are low-level details that may be skipped unless they want to known how their contracts are compiled. (I never saw a Java manual explaining what the programs look like in bytecode!)

NicNomadic: What about moving this section in the page for Unity, it it's only needed there? Or in a page a bit more general about web3 libraries for Tezos?

Well, I agree that it *marginally* impacts all the callers of contracts using octez-client, *if* they forget to specify an entry point. When you do indicate an entry point to octez-client, you just give the argument as that entry point expects it, no need to wrap it yourself as the default entry point would expect it, right?
If so, it's not so important why the call fails when no entry point is specified, users may just assume that they must specify an entry point, and we avoid all these hairy details.

NicNomadic: In other words, does this work:

    octez-client --wait none transfer 5 from tim to KT1R2LTg3mQoLvHtUjo2xSi7RMBUJ1sJkDiD --entrypoint "reset" --arg 'unit' --burn-cap 0.1

docs/smart-contracts/entrypoints.md
```diff
@@ -12,77 +12,66 @@ The entrypoints in a Tezos smart contract must meet these specifications:
...
+For an example of a simple contract, see the tutorial [Create a smart contract](../tutorials/smart-contract).
+
+For examples of more complex contracts, see [Examples of contracts](https://opentezos.com/smart-contracts/simplified-contracts/) on opentezos.com.
 
 ## Entrypoint logic
 
-An entrypoint may run all kinds of logic based on:
+An entrypoint may run logic based on:
 
 - Its storage
```
zamrokk, line 29: contract storage. An entrypoint does not have it sown storage, it is global to the contract

docs/smart-contracts/entrypoints.md
```diff
@@ -12,77 +12,66 @@ The entrypoints in a Tezos smart contract must meet these specifications:
...
 - The parameters that senders pass
 - Global values such as the address of the caller
```
zamrokk, line 31: No, it is not really global values but rather a Transaction context

docs/smart-contracts/entrypoints.md
```diff
@@ -12,77 +12,66 @@ The entrypoints in a Tezos smart contract must meet these specifications:
...
 - The table of constants
 <!-- TODO link to Tezos library for address of caller/globals and table of contents -->
 
-Entrypoints may not access information outside Tezos, such as calling external APIs.
+Entrypoints may not access information outside of Tezos, such as calling external APIs.
```
zamrokk, line 35: it is just impossible, I would replace it with "cannot"

docs/smart-contracts/entrypoints.md
```diff
@@ -12,77 +12,66 @@ The entrypoints in a Tezos smart contract must meet these specifications:
...
-When calling a smart contract, senders provide either the full parameter as a variant or specify the name of the entrypoint and the corresponding parameter value.
+Developers need to know about the default entrypoint only when they encode parameters for smart contracts manually.
+Most Tezos clients, including Octez and Taquito, encode parameters automatically.
+Working from the previous example, they convert a call to the `increment` entrypoint with the parameter 5 as a call to the default entrypoint with the parameter `Right (Right 5)`.
 
-Other languages have different ways of indicating entrypoints.
+Different languages have different ways of indicating entrypoints.
 For information about coding entrypoints in specific languages, see these links:
 
 - Michelson: [Entrypoint](https://tezos.gitlab.io/active/michelson.html#entrypoints)
```
zamrokk, line 77: maybe add a note on the octez-cli ? you can point to a specific entrypoint on arguments and just pass the expected payload without LEFT RIGHT stuff

docs/smart-contracts/entrypoints.md
timothymcmackin: It has to be a capital U in Unit, but yes, that works:
```
octez-client --wait none transfer 0 from tim to KT1R2LTg3mQoLvHtUjo2xSi7RMBUJ1sJkDiD --entrypoint "reset" --arg 'Unit' --burn-cap 0.1
```

docs/smart-contracts/entrypoints.md
timothymcmackin: Thanks -- updated.

docs/smart-contracts/entrypoints.md
timothymcmackin: Maybe this is a good example of what's happening here: These two calls are equivalent for the contract we're discussing:

```
octez-client --wait none transfer 0 from tim to KT1R2LTg3mQoLvHtUjo2xSi7RMBUJ1sJkDiD --arg 'Left Unit' --burn-cap 0.1
```
```
octez-client --wait none transfer 0 from tim to KT1R2LTg3mQoLvHtUjo2xSi7RMBUJ1sJkDiD --entrypoint "reset" --arg 'Unit' --burn-cap 0.1
```

docs/smart-contracts/entrypoints.md
timothymcmackin: Changed to "Transaction context values such as the address of the caller".

docs/smart-contracts/entrypoints.md
timothymcmackin: Yeah, octez may be a good example. How's this?

https://github.com/trilitech/tezos-developer-docs/pull/201/commits/daa013dc23add3692ba407e2ce8366ce2f34c695

docs/smart-contracts/entrypoints.md
NicNomadic: yes it's a good example, independently of where you explain it

NicNomadic: So if you leave it here, please start the section with a note that this section may be skipped at a first reading, or so.

NicNomadic: I mean, both sections "Implementation details: the default entry point" and "Entrypoint implementation". BTW, I don't see well the difference between the two, perhaps they should be collapsed.

timothymcmackin: Collapsed the sections to one section named "Implementation details: the default entrypoint."

Added this note that this info is very technical and can be ignored in most cases:

> In most cases, developers can ignore the default entrypoint and imagine that the compiled Michelson code has multiple entrypoints like the higher-level code. However, in some cases, you may need to consider how the contract actually decides which code to run and how clients trigger this code.

timothymcmackin:
@NicNomadic @zamrokk @rafoo Any further comments here?

PR #204, opened 2023-12-06
Info about counters for implicit accounts

timothymcmackin:
This may not be right right place for this very technical info.


timothymcmackin:
Based on discussions with development, counters are a very technical thing and not appropriate for this overview information. I'll think about where else it could go.

PR #205, opened 2023-12-07
Mocking up a larger doc structure

timothymcmackin:
Wondering what would happen if we pulled in the docs for the smart contract languages into docs.tezos.com:
https://docs-staging-git-consolidated-structure-mockup-trili-tech.vercel.app/smart-contracts/languages


timothymcmackin:
- Withdrawn. In the short term it's going to be better to get the language-specific sites organized in a consistent way and linking to each other properly than jamming them all onto one site. We may want to move smartpy to smartpy.tezos.com instead of jamming its into into docs.tezos.com.

PR #219, opened 2023-12-18
dApp quickstart

timothymcmackin:
Fast as possible walkthrough of setting up a dApp. Preview: https://docs-staging-git-dapp-quickstart-trili-tech.vercel.app/dApps/quickstart


zamrokk:
Not sure it is a good idea to propose a quickstart with some "cooked" code.

If there is no step by step approach from scratch, you hide the complexity of installing some dependencies (taquito, etc...) and specific configuration (like the polyfills issues) 

I fear that the reader wil not learn a lot, just runnning command to launch the app  :/

zamrokk:
Also it just talk about the frontend part without explaining the link with a deployed smart contract. It think it is important to explain how this link is done

timothymcmackin:
@zamrokk Ok, I've restored a version that walks you through setting up the app with svelte and copying the code in. It's basically a tutorial minus most of the explanation of what's going on.

I also figured  that because this is a dApp quickstart, it should use an existing smart contract and handle just setting up the web frontend. Later I can do a smart contract quickstart for the other side. What do you think?

beatalipska (approved):

docs/dApps/quickstart.md
```diff
@@ -1,5 +1,405 @@
-# dApp quickstart
```
NicNomadic, line 1: Globally, I have the impression that this quickstart dapp is very similar, and roughly the same order of complexity as the one in "Build your first dapp" (in https://docs-staging-git-dapp-quickstart-trili-tech.vercel.app/tutorials/build-your-first-app). Is it worth introducing this new howto? Maybe it would be better and simpler to leave the introducing docs in the docs section, and link for a quickstart directly to the other one?

timothymcmackin: Yes, after leaving this PR sit for a while I think it may not be worth duplicating the content in the tutorials. This quickstart just provided a tutorial with as little explanation as possible. Might not be worth the cost of adding and maintaining it.

NicNomadic: I see the initial motivation, but the result shows one cannot significantly reduce the amount of details. And it also blurs the distinction between the doc part and the tutorials part on the website.

So yes, let's drop it.

timothymcmackin:
Withdrawn. I'll add links from the landing page to the beginner-level tutorials and that's probably good enough as a quickstart.

PR #228, opened 2023-12-28
Set up CSP headers

timothymcmackin:
Need to get CSP headers working so we can use google analytics.


timothymcmackin:
Merging based on discussion with Craig.

PR #238, opened 2024-01-02
Update and clarify docs on indexers; remove indexer setup walkthroughs

timothymcmackin:
- Update info on indexers
- Remove the "self-study" part and make it a tutorial or an article later: https://github.com/trilitech/tezos-developer-docs/commit/99e0425c957f3aeb59c2d8320de36980ee0fad7b
- Remove topic on setting up custom indexers because that's covered on opentezos

We could consider removing the "Where indexers are used" and "Data available on indexers" sections, too.


docs/developing/information/indexers.md
```diff
@@ -1,429 +1,186 @@
...
+Indexers can organize data in custom ways.
+For example, TzKT indexes FA1.2 and FA2 tokens and gives each an internal ID.
+This ID can be faster and easier to work with and compare than the contract ID.
 
-The exact list of tables, indexes schemas, and command syntax depends on the indexer and database it uses. While using TzKT, a query for a USDT balance will look like this:
+This TzKT query gets information about an FA1.2 token based on its internal ID instead of its contract address:
 
 ```
-https://api.tzkt.io/v1/tokens/balances?token.contract=KT1XnTn74bUtxHfDtBmm2bGZAQfhPbvKWR8o&account=tz1a1RTsGUbads3VucUQDxJF4EDXkDWcDHPK
+https://staging.api.tzkt.io/v1/tokens?id=42290944933889
```
dmirgaleev, line 103: https://api.tzkt.io/v1/tokens?id=42290944933889 is better

docs/developing/information/indexers.md
```diff
@@ -1,429 +1,186 @@
...
 
-State updates happen at each block, which means all the balance updates are always verified, and the indexer will follow chain reorganizations in real-time.
+Usually, they are used in projects that require only specific on-chain data, such as active user balances, balances of their smart contracts, and NFT metadata.
+You can optimize a custom selective indexer for fast execution of specific project queries.
+You can also use them to work with data in the way that you want to, such as storing token amounts with decimal places.
 
+You can use these frameworks to create your own selective indexer to provide the data that you need in the format that you need:
 
-## A simple explanation of how indexers work
+- [Que Pasa](https://github.com/tzConnectBerlin/que-pasa)
```
dmirgaleev, line 66: The last commit for Que Pasa 2 year ago. Perhaps it's better to check if it works before.

docs/developing/information/indexers.md
timothymcmackin: Updated.

docs/developing/information/indexers.md
timothymcmackin: Ok, I'm going through these instructions on opentezos to see if I can get que pasa running:
https://opentezos.com/dapp/indexers/que_pasa/

timothymcmackin: I got que-pasa to work with some hacks:

- It doesn't like that some RPCs return 404s including the data, so I used https://rpc.tzbeta.net per your suggestion
- `--enable-bcd` doesn't seem to work (it throws timeout errors), so I didn't use it
- I had to manually go into config.rs and set NODE_URL and DATABASE_URL because it doesn't seem to respect environment variables

But after all that I can run it and see it populate the index with data about the contract.

PR #239, opened 2024-01-03
Add dry run for reset entrypoint

timothymcmackin:
Had a user get confused trying to dry-run the reset entrypoint, so we need to mention "unit."


timothymcmackin:
Adebola approved offline.

PR #249, opened 2024-01-05
Put GTM first so it gets loaded regardless of the ; problem

timothymcmackin:
We might not have to do the post-build replace if we put GTM first in the list of scripts to allow with CSP.


timothymcmackin:
Merging now to see if this works in staging.

PR #266, opened 2024-01-16
Rework testing topics

timothymcmackin:
- Move testing smart contracts topic into developing section
- Recast that topic as "Testing locally"
- Recast testnets topic as "Testing on sandboxes and testnets"
- Add info on how you have to use a specific docker image or build of octez in oder to use weeklynet and dailynet


beatalipska (dismissed):
all good, makes a lot of sense!

onedebos (approved):

PR #275, opened 2024-01-22
Clarify this note about methods

timothymcmackin:
I think I understand what this note is saying better now.


timothymcmackin:
Thomas approved offline.

PR #221, opened 2023-12-19
Simple landing page for dApps section

timothymcmackin:
Preview: https://docs-staging-git-dapp-landing-page-trili-tech.vercel.app/dApps

PR #287, opened 2024-01-29
Clarify dApp landing page

timothymcmackin:
Nic had some comments after https://github.com/trilitech/tezos-developer-docs/pull/221 was merged. This PR includes them.


NicNomadic:
Sorry, I don't see any difference in the preview at <https://docs-staging-git-nic-comments-on-dapp-landing-page-trili-tech.vercel.app/dApps>. Am I looking at the right place?

docs/dApps.mdx
```diff
@@ -17,31 +17,17 @@ In general, dApps have these parts:
 - **Backend**: An on-chain component that consists of one or more [smart contracts](./smart-contracts)
 
 The off-chain component can be nearly any kind of program, including a web application, mobile or desktop app, or command-line interface.
-It relies on wallets and tools to interact with the smart contract on behalf of a user's Tezos account.
+It relies on wallets and tools to interact with the smart contracts on behalf of a user's Tezos account.
 
-<LucidDiagram width="640px" height="480px" src="https://lucid.app/documents/embedded/8caf9ef1-11e4-454a-bbb6-ef4852515959" id=".4NXymECcQqz" />
+<LucidDiagram width="640px" height="360px" src="https://lucid.app/documents/embedded/8caf9ef1-11e4-454a-bbb6-ef4852515959" id=".4NXymECcQqz" />
```
NicNomadic, line 22: Referring to an artifact which is editable online separately from any github control seems a real flaw, isn't it? Why don't you just export your picture as a static image, and perhaps link in a comment to the source, in case one wants to edit it?

timothymcmackin: Ok, I put it in as a static image. In cases of large diagrams I like offering the LucidChart container because people can scroll and zoom to explore the diagram, but yes, it means that the artifact is not in the repository.

NicNomadic: It's much better like this, both from usability and security viewpoints.

People can zoom in without the Lucid source, isn't it? If you mean that they would have more resolution, maybe you can try to improve that in the static image, either increasing its resolution, or exporting SVG (which is supposed to be more scalable, as the name implies).

PR #307, opened 2024-02-06
Add indexers to architecture diagram

timothymcmackin:
- Add indexers to the architecture diagram
- Hard-code the diagram instead of linking to live diagram on lucid


beatalipska (approved):
boom!

PR #310, opened 2024-02-07
Need to use the same case for this constant

timothymcmackin:
On line 46 it is in all caps.


timothymcmackin:
Included in https://github.com/trilitech/tezos-developer-docs/pull/312.

PR #311, opened 2024-02-07
Clarify setting the slot in the publish command

timothymcmackin:
You have to select your slot in the following command, so let's make that clear to the reader.


timothymcmackin:
Included in https://github.com/trilitech/tezos-developer-docs/pull/312

PR #312, opened 2024-02-07
Clarify DAL file tutorial

timothymcmackin:
- Add troubleshooting info
- Clarify the process of re-deploying the Smart Rollup
- Add info on the response of the publishing command
- Hard code images instead of linking to LucidChart
- Move MDX files to MD because they are no longer using MDX features
- Based on user feedback, clarify in detail what you do and don't set up in the tutorial
- Add diagram of what you set up in the tutorial


timothymcmackin:
Merging with offline approvals

PR #321, opened 2024-02-14
DAL tutorial updates

timothymcmackin:
Updates and clarifications for the DAL file server tutorial:

- Clarify how to get a text editor and how to use docker volumes to make editing in the container easier
- Clarify some commands based on feedback
- How to hide the warning message


francoisthire:
I did not execute the commands added, but changes LGTM.

PR #328, opened 2024-02-16
[wip] Updated views topic

timothymcmackin:
Working on a topic  on views that would fit with the opentezos organization so we could reuse the content.

### DIFF
--- a/docs/smart-contracts/views.md
+++ b/docs/smart-contracts/views.md
@@ -17,68 +17,69 @@ The call to the view runs immediately and the return value can be used in the ne
 In particular, it doesn't modify the storage of its contract and doesn't generate any operations.
 - Views do not include the transfer of any tez.
 
-## Example View
+## Examples
 
-Here is an example that uses a view.
-The following contract is a ledger that handles a fungible token and keeps track of how many tokens are owned by each user.
+Views can provide information about tokens.
+For example, the [FA2 standard](../architecture/tokens/FA2) suggests that contracts provide several views, including a view that provides the amount of a token that a specified account owns.
 
-{<table>
-  <caption>Ledger contract</caption>
-  <thead>
-    <tr>
-      <th>Storage</th>
-      <th>Entrypoint effects</th>
-    </tr>
-  </thead>
-  <tbody>
-    <tr>
-      <td>
-        <ul>
-          <li><code>ledger</code>: <code>big-map</code>
-            <ul>
-              <li>Key:<ul>
-                  <li><code>user</code>: <code>address</code></li>
-                </ul>
-              </li>
-              <li>Value:<ul>
-                  <li><code>tokens</code>: <code>nat</code></li>
-                </ul>
-              </li>
-            </ul>
-          </li>
-        </ul>
-      </td>
-      <td>
-        <ul>
-          <li><code>view getBalance(user: address)</code>
-            <ul>
-              <li>return <code>ledger[user].tokens</code></li>
-            </ul>
-          </li>
-          <li><code>transfer(nbTokens, destination)</code>
-            <ul>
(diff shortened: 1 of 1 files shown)


docs/smart-contracts/views.md
```diff
@@ -2,86 +2,194 @@
...
+```bash
+octez-client run view "get_larger" on contract "KT1Uh4MjPoaiFbyJyv8TcsZVpsbE2fNm9VKX" with input "Pair 4 5"
 ```
 
+If the view takes no parameters, you can pass Unit or omit the `with input`.
+<!-- TODO link to info on encoding param values -->
+
 ## Implementation details
 
 - Michelson: [Operations on views](https://tezos.gitlab.io/active/michelson.html#operations-on-views)
```
NicNomadic, line 191: Possibly also refer to this standalone page, which is much shorter that this one, but contains useful context about other kinds of views: <https://tezos.gitlab.io/active/views.html>

NicNomadic: In fact, the link you give doesn't even work anymore, so you should replace it with the page I suggested, but mind changing the link description.

docs/smart-contracts/views.md
```diff
@@ -2,86 +2,194 @@
...
+Smart contracts can provide information via callbacks, but using a callback means calling entrypoints, which is an asynchronous action.
+
+By contrast, views are synchronous; a contract can call a view and use the information that it returns immediately.
+
+Like entrypoints, views can accept parameters, access the contract's storage, and call other views.
+Unlike entrypoints, views return a value directly to the caller.
+However, views can't cause side effects, so they can't create operations, including calling smart contracts and transferring tez.
+Views also can't change the contract storage.
+
+Nodes can run views without creating any operations, which lets off-chain users get a node to run a view and return the result immediately, as a convenient way to get information from a smart contract.
```
NicNomadic, line 20: I don't understand this phrase at all. Maybe you mean something like this?
```suggestion
As views run synchronously, off-chain users can ask a node to run a view and return the result immediately; this is a convenient way to get information from a smart contract.
```

docs/smart-contracts/views.md
```diff
@@ -2,86 +2,194 @@
...
+  return match(valOpt) {
+    when(Some(val)): val;
+    when(None): "";
+  }
+}
+```
+
+## Calling views in JsLIGO
+
+This JsLIGO code calls the `get_larger` view from the previous example by passing the target contract address, parameters, and view name to the `Tezos.call_vew()` function:
```
NicNomadic, line 68: The call example could be much more meaningful on the get view (instead of the dummy get_larger view), as it really gets some useful info from the contract.

docs/smart-contracts/views.md
```diff
@@ -2,86 +2,194 @@
...
+}
+```
+
+This view returns a value from a big-map in storage:
+
+```ts
+type storageType = big_map<string, string>;
+
+@view
+const get = (key: string, s: storageType): string => {
```
NicNomadic, line 57: "get" sounds too generic, maybe rename this to "get_balance" or so?

docs/smart-contracts/views.md
timothymcmackin: I'll work on a better example.

timothymcmackin:
> I would like to leave a new comment on this phrase but I cannot in github:
> 
> > Because views run synchronously, off-chain users can ask a node to run a view and return the result immediately; this is a convenient way to get information from a smart contract.
> 
> In fact, the sync model has nothing to do here. We just want to say that off-chain users can also use them and how:
> 
> > Off-chain users can also run a view by using the Octez client `run view` command; this is a convenient way to get information from a smart contract.

Updated to:

> Off-chain users can run a view without creating a transaction, which is a convenient way to get information from a smart contract. For example, you can use the Octez client `run view` command to run a view from the command line.

PR #342, opened 2024-02-28
No need to configure the client so early

timothymcmackin:
In the DAL baker tutorial, we initialize the client in step 1 but do not use it. Then in step 3 we set its configuration. It's simpler to do this config once, so this PR removes that client initialization in step 1.

### DIFF
--- a/docs/tutorials/join-dal-baker/prepare-account.md
+++ b/docs/tutorials/join-dal-baker/prepare-account.md
@@ -2,32 +2,17 @@
 title: "Step 3: Set up a baker account on Weeklynet"
 authors: Tezos core developers
 last_update:
-  date: 28 February 2024
+  date: 6 March 2024
 ---
 
 Our baker needs a user account consisting of a pair of keys and an address.
+In this section, you use the Octez client to create an account and register it as a delegate.
 
 1. Open a new terminal window in the same environment.
 If you are using a Docker container, you can enter the container with the `docker exec` command, as in `docker exec -it my-image /bin/sh`.
 To get the name of the Docker container, you run the `docker ps` command.
 
-1. Set your installation of the Octez client to use your node:
-
-   - If you have not used the Octez client yet, run this command to initialize it:
-
-      ```bash
-      octez-client -E http://127.0.0.1:8732 config init
-      ```
-
-   - If you have already initialized the Octez client, run this command to instruct it to query your node:
-
-      ```bash
-      octez-client -E http://127.0.0.1:8732 config update
-      ```
-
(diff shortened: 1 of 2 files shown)


docs/tutorials/join-dal-baker/prepare-account.md
```diff
@@ -13,12 +13,28 @@ To get the name of the Docker container, you run the `docker ps` command.
 
 1. Set your installation of the Octez client to use your node instead of the public Weeklynet node:
 
+   - If you have not used the Octez client yet, run this command to initialize it:
+
+   ```bash
+   octez-client -E http://127.0.0.1:8732 config init
+   ```
+
+   If you have already initialized the Octez client, run this command to switch to your node:
```
ezal, line 22: Maybe this a bit more precise:

```suggestion
   If you have already initialized the Octez client, run this command to instruct it to query your node:
```

rafoo:
The problem is not that the client is configured too early, the problem is that it is configured at all. The client configuration is used by the baking daemon so setting the default endpoint to anything but localhost is a bad idea for a baker setup.

timothymcmackin:
> The client configuration is used by the baking daemon

Oh, I didn't know that. Should the tutorial make that explicit?

rafoo:
IMO the simplest is to remove both the `octez-client config init` and `octez-client config update` lines and come back to explicit `octez-cilent -E blah` for the rare cases where we want an endpoint different from the localhost. If we are afraid that some user may have followed the instructions of the DAL+rollup tutorial (which, IIRC, also contains an `octez-client -E ... config init` line) we could advise to check that the configured endpoint is localhost (with `octez-client config show | jq .endpoint`).

rafoo:
Do you want to keep the `octez-client config init` in step 1? If you choose to remove it there are some `--endpoint <URL of the public node>` to add here and there because we may want to check the account balance and to register as a delegate before the localhost node is bootstrapped.

timothymcmackin:
> Do you want to keep the `octez-client config init` in step 1? If you choose to remove it there are some `--endpoint <URL of the public node>` to add here and there because we may want to check the account balance and to register as a delegate before the localhost node is bootstrapped.

This PR removes the `octez-client config init` in step 1:
https://docs-staging-git-remove-redundant-client-config-trili-tech.vercel.app/tutorials/join-dal-baker/get-octez

rafoo:
The instructions in step 3 should have no effect most of the time but it's not a problem.

rafoo (approved):

rafoo:
LGTM.

PR #347, opened 2024-03-01
Update tezos-different.md

NicNomadic:
Here are some changes resulting from a review of the page "What makes Tezos different?"

### DIFF
--- a/docs/overview/tezos-different.md
+++ b/docs/overview/tezos-different.md
@@ -8,9 +8,9 @@ Here are some of the features that make Tezos different from other blockchains:
 
 ## Tezos can upgrade itself
 
-Tezos has a built-in capability to upgrade itself, which allows the network to evolve without requiring a hard fork. Anyone can propose an upgrade to the protocol and have it adopted by the network without compromising the platform's stability or causing fragmentation. This feature allows Tezos to adapt to new technologies and to address user needs rapidly.
+Tezos has a built-in capability to upgrade itself, which allows the network to evolve without requiring a hard fork. Anyone can propose an upgrade to the protocol and have it adopted by the network without compromising the platform's stability or causing fragmentation. This feature allows Tezos to adapt regularly to new technologies and to address user needs rapidly.
 
-## Stakeholders participate in governance
+## Everyone can participate in governance
 
 Anyone who holds XTZ — the chain's native token — can propose changes to how Tezos works, such as changes to gas fees and block times, new features such as Smart Rollups, or even major changes like how the consensus mechanism works.
 
@@ -20,12 +20,12 @@ Anyone who holds XTZ — the chain's native token — can propose changes to how
 
 ## Tezos uses proof of stake
 
-The proof-of-stake consensus model eliminates the need for high energy use, making it the "green" choice for blockchains. Instead of competing to achieve consensus as in proof-of-work models, Tezos nodes (called *bakers*) stake Tezos tokens to earn the right to create blocks and receive rewards. Users who want to participate without running a node themselves can delegate tokens to a baker for a share of the rewards. The bakers and delegators keep control of their tokens and can remove them at any time. Tezos's approach to consensus has been described as [Liquid Proof of Stake](https://medium.com/tezos/liquid-proof-of-stake-aec2f7ef1da7).
+The proof-of-stake consensus model eliminates the need for high energy use, making it the "green" choice for blockchains. Instead of competing to achieve consensus as in proof-of-work models, Tezos nodes (called *bakers*) stake Tezos tokens to earn the right to create blocks and receive rewards. Users who want to participate without running a node themselves can delegate tokens to a baker for a share of the rewards. The bakers keep control of their tokens in that they can unstake them later; and delegators keep complete control of their tokens. Tezos's approach to consensus has been described as [Liquid Proof of Stake](https://medium.com/tezos/liquid-proof-of-stake-aec2f7ef1da7).
 
-The proof-of-stake model improves scalability and encourages incentive alignment. It also increases the cost of 51% attacks and avoids environmentally wasteful proof-of-work. Tezos launched in June 2018 as one of the first major proof-of-stake networks.
+The proof-of-stake model improves scalability and encourages cooperation via incentives. It also increases the cost of 51% attacks and avoids environmentally wasteful proof-of-work. Tezos launched in June 2018 as one of the first major proof-of-stake networks.
 
 For more information about how Tezos handles proof of stake, see https://tezos.gitlab.io/alpha/proof_of_stake.html.
 
 ## Tezos accepts multiple languages
 
-Tezos lets developers use languages that make sense for their use case, including versions of Python and JavaScript/TypeScript. For more information, see [Languages](../smart-contracts/languages/).
+Tezos provides a few different languages for developers to choose from, according to their use case, including versions of Python and JavaScript/TypeScript. For more information, see [Languages](../smart-contracts/languages/).


docs/overview/tezos-different.md
```diff
@@ -20,12 +20,12 @@ Anyone who holds XTZ — the chain's native token — can propose changes to how
...
+The proof-of-stake consensus model eliminates the need for high energy use, making it the "green" choice for blockchains. Instead of competing to achieve consensus as in proof-of-work models, Tezos nodes (called *bakers*) stake Tezos tokens to earn the right to create blocks and receive rewards. Users who want to participate without running a node themselves can delegate tokens to a baker for a share of the rewards. The bakers and delegators keep control of their tokens: bakers can vary their staked amount in time, and delegators can spend their tokens or change their delegate at any time. Tezos's approach to consensus has been described as [Liquid Proof of Stake](https://medium.com/tezos/liquid-proof-of-stake-aec2f7ef1da7).
 
 The proof-of-stake model improves scalability and encourages incentive alignment. It also increases the cost of 51% attacks and avoids environmentally wasteful proof-of-work. Tezos launched in June 2018 as one of the first major proof-of-stake networks.
```
NicNomadic, line 25: I find that "encourages incentive alignment" is difficult to understand. Do you mean that incentives are designed to be aligned for the bakers (and delegators)? If so, it doesn't "encourage" but rather 'aims at" that, isn't it?

docs/overview/tezos-different.md
```diff
@@ -20,12 +20,12 @@ Anyone who holds XTZ — the chain's native token — can propose changes to how
...
 
 For more information about how Tezos handles proof of stake, see https://tezos.gitlab.io/alpha/proof_of_stake.html.
 
 ## Tezos accepts multiple languages
 
-Tezos lets developers use languages that make sense for their use case, including versions of Python and JavaScript/TypeScript. For more information, see [Languages](../smart-contracts/languages/).
\ No newline at end of file
+Tezos lets developers use languages that make sense for their use case, including versions of Python and JavaScript/TypeScript. For more information, see [Languages](../smart-contracts/languages/).
```
NicNomadic, line 31: Maybe state that there is not an unlimited subset, but just a few available languages to choose from?
```suggestion
Tezos provides a few different languages for developers to choose from, according to their use case, including versions of Python and JavaScript/TypeScript. For more information, see [Languages](../smart-contracts/languages/).
```

NicNomadic:
The title section "Stakeholders participate in governance" doesn't convey the key idea, because this phrase is true in any blockchain, if by "stakeholders" you understand influential core developers, for instance. What would be more specific for Tezos is something like "Everyone can participate in governance" or "Any user can participate in governance".

NicNomadic:
(I cannot add a comment on the unmodified lines so I put here more comments)

I would add the word "regularly" in the phrase "This feature allows Tezos to [regularly] adapt to new technologies", to express the fact that Tezos not only "can upgrade itself", but really does it on a regular basis.

timothymcmackin:
> The page https://opentezos.com/formal-verification doesn't seem to work correctly, should we change this link?

That's the correct link but there seems to be a problem with the page. Sometimes I see a blank page but when I switch to mobile layout by changing the browser dimensions it reappears. Can't reproduce it locally. I sent a message to Mathias about it.

docs/overview/tezos-different.md
```diff
@@ -20,12 +20,12 @@ Anyone who holds XTZ — the chain's native token — can propose changes to how
...
+The proof-of-stake consensus model eliminates the need for high energy use, making it the "green" choice for blockchains. Instead of competing to achieve consensus as in proof-of-work models, Tezos nodes (called *bakers*) stake Tezos tokens to earn the right to create blocks and receive rewards. Users who want to participate without running a node themselves can delegate tokens to a baker for a share of the rewards. The bakers and delegators keep control of their tokens: bakers can vary their staked amount in time, and delegators can spend their tokens or change their delegate at any time. Tezos's approach to consensus has been described as [Liquid Proof of Stake](https://medium.com/tezos/liquid-proof-of-stake-aec2f7ef1da7).
```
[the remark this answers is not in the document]
timothymcmackin: Even though "can remove them at any time" is not technically accurate, your version doesn't say that bakers can withdraw their tokens at all. How about "The bakers and delegators keep control of their tokens and can withdraw them later." This should be enough at  this level of detail.

docs/overview/tezos-different.md
timothymcmackin: Not sure what this is meant to mean. Perhaps it means that proof-of-work is a competitive model while the delegated proof of stake model allows users to pool resources and work together? Maybe it also means that the whole exercise is a collaborative effort between all bakers and delegators.

timothymcmackin:
> (I cannot add a comment on the unmodified lines so I put here more comments)
> 
> I would add the word "regularly" in the phrase "This feature allows Tezos to [regularly] adapt to new technologies", to express the fact that Tezos not only "can upgrade itself", but really does it on a regular basis.

Sure, that makes sense. Maybe it could be "regularly and rapidly" to illustrate that the update process happens relatively quickly.

timothymcmackin (changes requested):

docs/overview/tezos-different.md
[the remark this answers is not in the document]
NicNomadic: OK for your version, it removes the mistake, while staying more hihg-level.

docs/overview/tezos-different.md
NicNomadic: OK, the what about being slightly more explicit about it, like:
```suggestion
The proof-of-stake model improves scalability and encourages cooperation through incentives alignment. It also increases the cost of 51% attacks and avoids environmentally wasteful proof-of-work. Tezos launched in June 2018 as one of the first major proof-of-stake networks.
```
(so let's avoid explicitly comparing to POW, because it seems that POW, beyond Bitcoin, can also be fixed for more incentives alignment, see <https://cs251.stanford.edu/lectures/lecture6.pdf>)

docs/overview/tezos-different.md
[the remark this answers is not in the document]
NicNomadic: In fact, let's be a bit more coherent, opposing "stake" with "unstake", instead of "withdraw", and also suggest that the control is not total, but is about unstaking later:
```suggestion
The proof-of-stake consensus model eliminates the need for high energy use, making it the "green" choice for blockchains. Instead of competing to achieve consensus as in proof-of-work models, Tezos nodes (called *bakers*) stake Tezos tokens to earn the right to create blocks and receive rewards. Users who want to participate without running a node themselves can delegate tokens to a baker for a share of the rewards. The bakers and delegators keep control of their tokens in that they can unstake them later. Tezos's approach to consensus has been described as [Liquid Proof of Stake](https://medium.com/tezos/liquid-proof-of-stake-aec2f7ef1da7).
```

NicNomadic:
> > The page https://opentezos.com/formal-verification doesn't seem to work correctly, should we change this link?
> 
> That's the correct link but there seems to be a problem with the page. Sometimes I see a blank page but when I switch to mobile layout by changing the browser dimensions it reappears. Can't reproduce it locally. I sent a message to Mathias about it.

Solved by Mathias today

NicNomadic:
> > (I cannot add a comment on the unmodified lines so I put here more comments)
> > I would add the word "regularly" in the phrase "This feature allows Tezos to [regularly] adapt to new technologies", to express the fact that Tezos not only "can upgrade itself", but really does it on a regular basis.
> 
> Sure, that makes sense. Maybe it could be "regularly and rapidly" to illustrate that the update process happens relatively quickly.

Done (didn't add "rapidly" because there was already a "rapidly" in that phrase)

docs/overview/tezos-different.md
timothymcmackin: I still think "incentives alignment" is cryptic. What about "encourages users to pool resources?"

timothymcmackin:
> > The page https://opentezos.com/formal-verification doesn't seem to work correctly, should we change this link?
> 
> That's the correct link but there seems to be a problem with the page. Sometimes I see a blank page but when I switch to mobile layout by changing the browser dimensions it reappears. Can't reproduce it locally. I sent a message to Mathias about it.

This is a technical problem with opentezos that will be fixed soon.

docs/overview/tezos-different.md
NicNomadic: I do find this quite intuitive (once I googled around for it), byt it's true that I cannot find an accessible and neutral reference for this term. Then what about just "and encourages cooperation through incentives"? Just staying general (cooperation), and everyone knows what's an incentive.

(we could also add "well-designed incentives", but that's optional)

timothymcmackin: "and encourages cooperation via incentives" sounds good to me.

NicNomadic: Done.

NicNomadic:
Are we ready for merging now?

PR #352, opened 2024-03-05
Set the module name instead of relying on the default

timothymcmackin:
The smart contract tutorial uses the namespace/module  `Counter`. Conveniently, this is the default module name in the IDE, but we should still make sure that it is set correctly.

Screenshot 2024-03-11 at 10 53 18 AM


docs/tutorials/smart-contract/jsligo.mdx
```diff
@@ -204,6 +204,12 @@ namespace Counter {
 
 Before you can deploy the contract to Tezos, you must compile it to Michelson, the base language of Tezos contracts.
 
+1. Set the compiler to compile the namespace in your code:
```
Laucans, line 207: I would say : 
"to target the namespace to compile in your code

docs/tutorials/smart-contract/cameligo.mdx
```diff
@@ -189,6 +189,12 @@ type returnValue = operation list * storage
 
 Before you can deploy the contract to Tezos, you must compile it to Michelson, the base language of Tezos contracts.
 
+1. Set the compiler to compile the namespace in your code:
```
Laucans, line 192: Namespace is module in cameligo. 

Also I would say "to target the module to compile in your code"

docs/tutorials/smart-contract/jsligo.mdx
timothymcmackin: That's better, thank you. Updated/

beatalipska (dismissed):
🚀

PR #371, opened 2024-04-02
[do not merge] Mockup of concepts section

timothymcmackin:
Considering adding a Concepts section to the Tezos docs. Not sure if it makes sense to add this kind of information here when opentezos.com is available, but people are making the case for some basic info on how blockchains work to be added here.


timothymcmackin:
Stale; closing.

PR #379, opened 2024-04-12
Update diagram to simplify text in blocks

timothymcmackin:
- Add a line from wallets to indexers because  they use indexers to get token balances.
- Remove examples of operations from blocks and just say "Transactions and other operations."


beatalipska (approved):
makes sense 👍🏻

PR #381, opened 2024-04-16
Simplify operations in main arch diagram

timothymcmackin:
Simplify this diagram by saying "transactions and other operations" instead of providing random examples of operations.

Before:

Screenshot 2024-04-16 at 9 17 46 AM

After:

Screenshot 2024-04-16 at 9 18 05 AM


beatalipska (approved):
boom 🎉

PR #383, opened 2024-04-17
Details about packing and unpacking

timothymcmackin:
Info about packing and unpacking and some gotchas about the formatting.

Preview: https://docs-staging-git-packing-details-trili-tech.vercel.app/smart-contracts/serialization#formatting

- Should the Octez docs mention this formatting and provide more detail, such as the possible values for prefixes?
- Is this info about the octez-codec useful here? I added just the command to print out all the formatting info. https://tezos.stackexchange.com/a/4000/10849

Source: https://tezos.stackexchange.com/questions/5879/conversion-of-address-to-bytes/5895#5895

### DIFF
--- a/docs/smart-contracts/serialization.md
+++ b/docs/smart-contracts/serialization.md
@@ -31,7 +31,6 @@ You can do so by packing it first and then applying a hash function such as `BLA
 
 ## Implementation details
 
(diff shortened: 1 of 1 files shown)


beatalipska (approved):
🚀

PR #361, opened 2024-03-18
Development landing page

timothymcmackin:
This is a simple container page but I think it might help developers get oriented to the tooling ecosystem.

PR #393, opened 2024-05-10
Fix card `Development environments` in the landing page

huancheng-trili:
Fix the link to a page that was moved in #361.

What can be more frustrating for a new developer than clicking into `start developing` just to see nothing but `page not found`? :)


timothymcmackin (approved):
Thanks so much for the fix!

PR #401, opened 2024-05-29
Info about the unit type

timothymcmackin:
Got some questions in discord about unit so I thought we should mention it here.

### DIFF
--- a/docs/smart-contracts/data-types/primitive-data-types.md
+++ b/docs/smart-contracts/data-types/primitive-data-types.md
@@ -2,7 +2,7 @@
 title: Primitive data types
 authors: 'Mathias Hiron (Nomadic Labs), Sasha Aldrick (TriliTech), Tim McMackin (TriliTech)'
 last_update:
-  date: 4 October 2023
+  date: 4 June 2024
 ---
 
 Tezos contracts support these primitive data types.
@@ -15,7 +15,6 @@ The high-level languages may implement these data types slightly differently, bu
 - [Booleans](#booleans)
 - [Timestamps](#timestamps)
 - [Addresses](#addresses)
-- [Unit](#unit)
 
 ## Numeric data types: `int` and `nat` {#numeric}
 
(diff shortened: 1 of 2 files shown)


docs/smart-contracts/data-types/primitive-data-types.md
```diff
@@ -15,6 +15,7 @@ The high-level languages may implement these data types slightly differently, bu
 - [Booleans](#booleans)
```
rafoo, line 15: (sorry I cannot comment above this line)

Shouldn't the last_update date be changed?

docs/smart-contracts/data-types/primitive-data-types.md
```diff
@@ -173,11 +174,29 @@ For smart contract accounts, the string starts with "KT1".
 
 The next part of the string is a `Base58` encoded hash, followed by a 4-byte checksum.
 
+## Unit {#unit}
+
+In Tezos, the `unit` type represents no value.
```
rafoo, line 179: This could be confused with the never type. There is a value of type unit but there is only one.

I propose:
> In Tezos, a value of type `unit` holds no information.

rafoo:
On the page in general, I would add a note at the beginning explaining that the list of types is not meant to be exhaustive and pointing to some exhaustive listings in the documentations of the various languages (for Michelson it would be https://tezos.gitlab.io/michelson-reference/#types).

docs/smart-contracts/data-types/primitive-data-types.md
timothymcmackin: In that case I'll move it to the topic on complex types because it's really a variant that contains no info, right? I can handle it like the LIGO docs:
https://ligolang.org/docs/variants/unit/?lang=jsligo

timothymcmackin:
> On the page in general, I would add a note at the beginning explaining that the list of types is not meant to be exhaustive and pointing to some exhaustive listings in the documentations of the various languages (for Michelson it would be https://tezos.gitlab.io/michelson-reference/#types).

Added a note about where to get info about all of the types that are available in each language.

docs/smart-contracts/data-types/primitive-data-types.md
timothymcmackin: Changed the description to:

> In Tezos, the `unit` type is a special case of the variant type: a single value that holds no information. It is a placeholder where a variable is required but no other information is needed.

rafoo: - A type is not a value.
- The concept of variant type is more complex than the concept of the unit type so defining the unit type as a special case of a variant type is does not very helpful, especially if the page does not define variant types.

So, new proposition:

> In Tezos, the `unit` type is a type containing a single value. It is used as a placeholder where some type is required but no information is transmitted. It is the input type of functions taking no input, the output type of functions producing no output, and the storage type of contracts storing no information.

rafoo: Sorry, I did not realize that this was moved to a page talking about record and variants.

So if you want to mention the relationship with variants and records, the `unit` type is both a variant with a single constructor and a record with no field. But I'm still not convinced that this is worth mentioning.

PR #407, opened 2024-06-10
DAL is now available on weeklynet

timothymcmackin:
Update docs to remove info about DAL not being available on Mainnet.

I'll look into updating the tutorials to use Ghostnet instead of Weeklynet, but it may depend on how long it takes to get baking rights on Ghostnet.


beatalipska (approved):
🤙🏻

PR #408, opened 2024-06-10
Info about protocol constants

timothymcmackin:
There's a bit more to protocol constants than this, such as the difference between fixed and parametric constants, but this seemed like the right level of detail for this page.


beatalipska (approved):
🎸

PR #410, opened 2024-06-10
Add DAL to the list of info sources for Smart Rollups

timothymcmackin:
Smart Rollups have access to three sources of data; mention the DAL where it wasn't mentioned before.


beatalipska (dismissed):
🚀

PR #411, opened 2024-06-10
Show the math for the length of the refutation period

timothymcmackin:
I worked out the math like this:

Screenshot 2024-06-11 at 7 51 09 AM


beatalipska (approved):
🚀

PR #422, opened 2024-07-11
Docs for FA2.1 standard

timothymcmackin:
Docs for FA2.1 in the same style as FA2.

Preview: https://docs-staging-git-fa-21-trili-tech.vercel.app/architecture/tokens/FA2.1


docs/architecture/tokens/FA2.1.md
Lfelli-nl, line 1: Actually this may be too confusing as this isn't the default case

Lfelli-nl (approved):

PR #424, opened 2024-07-24
[WIP] Update tutorial for Ghostnet

timothymcmackin:


docs/tutorials/build-files-archive-with-dal/using-full-slot.md
```diff
@@ -1,8 +1,8 @@
 ---
 title: "Part 4: Using the entire slot"
```
ezal, line 2:
```suggestion
title: "Part 5: Using the entire slot"
```

docs/tutorials/build-files-archive-with-dal/publishing-on-the-dal.md
```diff
@@ -1,14 +1,14 @@
 ---
 title: "Part 3: Publishing on the DAL"
```
ezal, line 2:
```suggestion
title: "Part 4: Publishing on the DAL"
```

ezal (approved):
The changes look ok to me.

I haven't been able to finish the tutorial, as expected:
* `latest` doesn't yet work with the current bootstrap node on Ghostnet
* slots are not attested because there are not enough DAL-enabled bakers on Ghostnet.

But I guess this PR is a preparation for the case when these two issues will be fixed.

docs/tutorials/build-files-archive-with-dal/using-full-slot.md
timothymcmackin: I've corrected the step numbering, thanks for pointing that out.

timothymcmackin:
I followed this tutorial on 27 January and it seems to work on Ghostnet. Can anyone else review? I think this update may finally be ready to go. @ezal @francoisthire

docs/tutorials/build-files-archive-with-dal/get-dal-params.md
```diff
@@ -167,25 +173,26 @@ cp target/wasm32-unknown-unknown/release/files_archive.wasm .
...
-To use it, save it in a file with an `sh` extension, such as `deploy_smart_rollup.sh` and give it executable permission.
+To use it, save it in a file with an `sh` extension, such as `deploy_smart_rollup.sh` and give it executable permission by running `chmod +x deploy_smart_rollup.sh`.
 Then you can run it any tme you update the `lib.rs` or `Cargo.toml` files to deploy a new Smart Rollup by passing your account alias, as in this example:
 
 ```bash
-./deploy_smart_rollup.sh $MY_ACCOUNT
+./deploy_smart_rollup.sh my_wallet
 ```
 
+This script assumes that your local node is running at http://127.0.0.1:8732.
```
francoisthire, line 191: I think the `octez-client` command above also requires this assumption. I suggest you move it before the first command that requires it.

docs/tutorials/build-files-archive-with-dal/get-slot-info.md
```diff
@@ -139,11 +155,10 @@ Follow these steps to update the Smart Rollup to access information about slot 0
    smart-rollup-installer get-reveal-installer -P _rollup_node/wasm_2_0_0 \
      -u files_archive.wasm -o installer.hex
 
-   octez-client --endpoint ${ENDPOINT} \
-     originate smart rollup files_archive from "${MY_ACCOUNT}" of kind wasm_2_0_0 \
+   octez-client originate smart rollup files_archive from my_wallet of kind wasm_2_0_0 \
      of type unit with kernel "$(cat installer.hex)" --burn-cap 2.0 --force
 
-   octez-smart-rollup-node --endpoint ${ENDPOINT} \
+   octez-smart-rollup-node --endpoint http://127.0.0.1:8732  \
```
francoisthire, line 161: Nit: I don't understand why sometime the endpoint is specified sometimes it is not. My guess would be than when it is not needed as for `octez-client`, you don't specify it, but for the `octez-smart-rollup-node` there might be no default so it needs to be specified?

Why not putting the endpoint everywhere so that a reader can copy/paste it easily and modify it if it uses a different endpoint?

francoisthire (approved):

docs/tutorials/build-files-archive-with-dal/get-dal-params.md
timothymcmackin: Added this:

>   The Octez client assumes that your local node is running at http://127.0.0.1:8732. If your node is running at a different host name or port, pass the host name and port of the node to the `--endpoint` argument. For example, if the node is running on port 8733, include `--endpoint http://127.0.0.1:8733` in the command.

docs/tutorials/build-files-archive-with-dal/get-slot-info.md
timothymcmackin: It's klunky and unnecessary to put the endpoint in every command. I'll remove the rest of them unless you really think people should specify --endpoint every time.

saroupille: I am fine with it.

PR #429, opened 2024-08-01
Update Smart Rollup tutorial

timothymcmackin:
Tutorial needs some updates, primarily that for the binary you can use `octez-smart-rollup-node` without the `alpha`.

Related issue: if you wait too long between origination and starting the Smart Rollup node, it can't get enough info from the rolling L1 node. This fix changes the sandbox for this tutorial to use an archive node: https://gitlab.com/trili/hello-world-kernel/-/merge_requests/4

### DIFF
--- a/docs/tutorials/smart-rollup/set-up.md
+++ b/docs/tutorials/smart-rollup/set-up.md
@@ -65,7 +65,7 @@ Later, you will use this image to run a sandbox Tezos environment for testing th
 1. Run this command to start the Docker image, open a command-line terminal in that image, and mount the `hello-world-kernel` folder in it:
 
    ```bash
-   docker run -it --rm --volume .:/home/tezos/hello-world-kernel --entrypoint /bin/sh --name octez-container tezos/tezos:master
+   docker run -it --rm --volume $(pwd):/home/tezos/hello-world-kernel --entrypoint /bin/sh --name octez-container tezos/tezos:master
    ```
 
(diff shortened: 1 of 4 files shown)


docs/tutorials/smart-rollup/set-up.md
```diff
@@ -65,7 +65,7 @@ Later, you will use this image to run a sandbox Tezos environment for testing th
 1. Run this command to start the Docker image, open a command-line terminal in that image, and mount the `hello-world-kernel` folder in it:
 
    ```bash
-   docker run -it --rm --volume $(pwd):/home/tezos/hello-world-kernel --entrypoint /bin/sh --name octez-container tezos/tezos:master
+   docker run -it --rm --volume .:/home/tezos/hello-world-kernel --entrypoint /bin/sh --name octez-container tezos/tezos:master
```
sribst, line 68: I don't know about this. `.` is a relative path where `${pwd}` gives the absolute path, which is prefered in general.

sribst (approved):

docs/tutorials/smart-rollup/set-up.md
timothymcmackin: Changed this back to `pwd`.

mebsout:
This file also needs the `-alpha` change I believe https://github.com/trilitech/tezos-developer-docs/blob/staging/docs/tutorials/smart-rollup/set-up.md#L80. Otherwise LGTM.

PR #450, opened 2024-09-12
Fix links

timothymcmackin:
Fix some broken links that came up with a link check.


NicNomadic:
The new links look good.

PR #471, opened 2024-11-18
Script to concatenate MD files

timothymcmackin:
This script concatenates all MD files and strips the markdown code into a single text file so you can feed it to an LLM. Run it by running `npm run concat`. It outputs to `allPageSourceFiles.txt` and if I've done it correctly, that file will be at https://docs.tezos.com/allPageSourceFiles.txt.

### DIFF
--- a/.github/workflows/deploy-production.yml
+++ b/.github/workflows/deploy-production.yml
@@ -20,7 +20,7 @@ jobs:
       - run: npm ci
       - run: npm run build
       - run: npm run concat
-      - run: mv allMdFilesConcatenated.md ./build/
+      - run: mv allPageSourceFiles.md ./build/
       - uses: aws-actions/configure-aws-credentials@v2
         with:
           aws-access-key-id: ${{ secrets.AWS_S3_PRODUCTION_KEY_ID }}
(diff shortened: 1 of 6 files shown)


beatalipska (approved):
magic!

NicNomadic:
I think it's mostly fine, except that you should skip the glossary, which is included as raw HTML, which is not very useful, because unreadable. Anyways, there's a readable copy of it in the Octez doc.

timothymcmackin:
> I think it's mostly fine, except that you should skip the glossary, which is included as raw HTML, so not very useful because unreadable. Anyways, there's a readable copy of it in the Octez doc.

Not sure if we should include it or not. LLMs may be able to read it as it is. I guess it depends on whether you're using this site alone or together with the Octez docs.

NicNomadic:
> Not sure if we should include it or not. LLMs may be able to read it as it is. I guess it depends on whether you're using this site alone or together with the Octez docs.

I'm pretty sure they don't interpret it the right way, because the HTML structure doesn't even expose it as a glossary, you just have some indented words and paragraphs, not marked as terms and definitions. But it's not a big deal if we leave it in, it's pretty small compared to other pieces of code (JS, HTML) inlined in the tutorials.

NicNomadic (dismissed):
LGTM

timothymcmackin:
Added filters for glossary and tutorials section.

timothymcmackin:
I changed this script to strip the markdown out so it would be plain text -- probably easier for the LLM to read.

beatalipska (approved):

Issue #473, opened 2024-11-21
Explicit the need and specificity of the duplicated tutorials for web-app-nft-minting


NicNomadic:
There are two tutorials for building a web app that mints NFTs:
- one in the Beginner section: https://docs.tezos.com/tutorials/create-an-nft/nft-web-app
- one in the Advanced section: https://docs.tezos.com/tutorials/create-an-nft/nft-taquito

Is the duplication justified?
- If yes, they should just  cross-reference each other in their intro, clearly explaining how they differ.
- If not, they should be fused together, or differentiated more.

timothymcmackin:
The one in the beginner section covers just the offchain application and uses a predeployed contract. It's much simpler; it uses only a single frontend application and does not require a backend application. The one in the intermediate section has the user deploy a contract and run a backend and frontend application.

So they cover the same subject matter at different levels of detail. If we're trying to trim overlap, I think the thing to do is to break up the intermediate tutorial into sections to make it easier to follow and move it to the beginner section. We could also remove its backed app, but I think we should keep it to make it more realistic for enterprise developers.

I'd like to keep at least one of these tutorials in the beginner section because we need some beginner tutorials that call contacts from web apps and the existing tutorial that does this (Start with a minimum dApp) is difficult to follow and covers a lot of different topics like taqueria and tickets.

Other opinions are welcome.

NicNomadic:
I have no objection for keeping both if they are so different. My point is: just explain with the same clarity in both tutorials what they cover and what the other tutorial does.

timothymcmackin:
Here's a draft of what it would look like with some info at the top of each  tutorial that explains what it covers and does not cover and what other NFT tutorials are available:
https://github.com/trilitech/tezos-developer-docs/pull/478

Not sure what is the best thing to do; very open to opinions.

timothymcmackin:
Related: Here is a PR to remove the tznft tutorial, which has overlap with the other  two NFT tutorials.

PR #478, opened 2024-11-28
Disambiguate NFT tutorials

timothymcmackin:
I created this as a potential solution to https://github.com/trilitech/tezos-developer-docs/issues/473.

Provides disambiguation for the 3 NFT-related tutorials. Not sure if it's better to do this or to reduce the number of tutorials. Very open to opinions. I did not mention the tznft tutorial because I'm considering removing it because tznft is no longer maintained and it's a very niche tool.

Fixes https://github.com/trilitech/tezos-developer-docs/issues/473.


NicNomadic (approved):
I think this small patch solves the issues in a quick way.

However, it would be useful to fuse them into a single tutorial. Probably the best way would be to:
- keep the complete tutorial
- scan the simple tutorial to import things that are missing from the first one, or are better explained here, if any
- include a shortcut in the complete tutorial to use the existing contract, if possible and not too disturbing

timothymcmackin:
Merging this for now. Made a note to look at refactoring the tutorials to merge them.

PR #490, opened 2024-12-13
Enable pushfeedback widget

timothymcmackin:
Enable the feedback widget. We've been using it on Etherlink's docs and it works well.
https://github.com/etherlinkcom/docs/pull/98


NicNomadic (approved):
LGTM. I noticed that on the landing page, the icon is behind that for the Bot, but I don't thinks it's a problem (and I don't know if this can be changed easily)

PR #495, opened 2024-12-18
Recast this as a node/baker/DAL tutorial

timothymcmackin:
This tutorial is going to be the primary place we point people to for instructions about running a node:
- Recast it from a DAL node tutorial to a node tutorial that includes a DAL node.
- Make consistent the systemd service files
- Use the `tezos` user to run the services

### DIFF
--- a/docs/tutorials/join-dal-baker.md
+++ b/docs/tutorials/join-dal-baker.md
@@ -1,14 +1,27 @@
 ---
-title: Join the DAL as a baker in 5 steps
+title: Run a Tezos node in 5 steps
 authors: Tezos core developers, Tim McMackin
 last_update:
-  date: 21 October 2024
+  date: 18 December 2024
 ---
+
+As described in [Nodes](/architecutre/nodes), Tezos nodes are peer-to-peer programs that run the Tezos network.
+Anyone can run a node, and they might do so for many different reasons, including:
+
+- Running nodes makes the Tezos network resilient and secure
+- Public nodes may have rate limits, so running your own node allows you to send unlimited requests to it to get information about Tezos or to send transactions from your dApps
+- Running a node is part of being a baker and receiving the rewards for baking
+
+This tutorial covers setting up a Tezos node as a baker, which includes running these processes:
+
+- An Octez node, sometimes referred to as a Tezos node
+- A baker daemon
+- A [Data Availability Layer](/architecture/data-availability-layer) node
+
+## Why is a DAL node needed?
 
 The Tezos data availability layer (DAL) is a peer-to-peer network that Tezos Smart Rollups can use to fetch data securely.
 The DAL is a key component for the scalability and bandwidth of Tezos and it's important for bakers to run DAL nodes along with their layer 1 nodes.
-
-## Why are DAL bakers needed?
 
 When users and dApps submit data to the DAL, bakers use DAL nodes to verify that the data is available.
 Then the bakers attest that the data is available.
--- a/docs/tutorials/join-dal-baker.md
+++ b/docs/tutorials/join-dal-baker.md
@@ -16,7 +16,7 @@ This tutorial covers setting up a Tezos node as a baker, which includes running 
 
 - An Octez node, sometimes referred to as a Tezos node
 - A baker daemon
-- A [Data Availability Layer](/architecture/data-availability-layer) node
+- A [DAL](/architecture/data-availability-layer) node
 
 ## Why is a DAL node needed?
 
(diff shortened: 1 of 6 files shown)


docs/tutorials/join-dal-baker.md
```diff
@@ -1,15 +1,28 @@
...
+
+- Running nodes makes the Tezos network resilient and secure
+- Public nodes may have rate limits, so running your own node allows you to send unlimited requests to it to get information about Tezos or to send transactions from your dApps
+- Running a node is part of being a baker and receiving the rewards for baking
+
+This tutorial covers setting up a Tezos node as a baker, which includes running these processes:
+
+- An Octez node, sometimes referred to as a Tezos node
+- A baker daemon
+- A [Data Availability Layer](/architecture/data-availability-layer) node
```
NicNomadic, line 19: As DAL is explained just below, you may shorten it here:
```suggestion
- A [DAL](/architecture/data-availability-layer) node
```

docs/tutorials/join-dal-baker/run-baker.md
```diff
@@ -19,6 +20,30 @@ Now that you have a DAL node, you can run a baking daemon that can attest to DAL
 Look up how to run programs persistently in the documentation for your operating system.
 You can also refer to [Run a persistent baking node](https://opentezos.com/node-baking/baking/persistent-baker/) on opentezos.com.
 
+   For example, if your operating system uses the `systemd` software suite, your service file might look like this example:
+
+   ```systemd
+   [Unit]
+   Description=Octez baker
+   Wants = network-online.target
```
NicNomadic, line 28: why are certain assignments spaced and others are not? Couldn't we be consistent, for instance, never spaced?

NicNomadic:
Maybe we should also run an accuser then, to be complete?

NicNomadic:
Also, we should just mention that it could be recommended to also run a signer, just referring to the doc for that (e.g. <https://tezos.gitlab.io/user/key-management.html#signer>).

NicNomadic (commented):
I like the idea to also include the node in the picture, so we can go ahead and complete the scene

docs/tutorials/join-dal-baker/run-baker.md
timothymcmackin: Made the spacing consistent.

timothymcmackin:
- Added optional steps for an accuser
- Mentioned remote signing
- Made the service files consistent

docs/tutorials/join-dal-baker/run-baker.md
```diff
@@ -99,3 +130,34 @@ Therefore, staking more tez brings more rewards but does not reduce the attestat
 - Attach DAL attestations: `ready to attach DAL attestation ...`
 
 Whether these messages appear or not after the attestation delay, proceed to [Step 5: Verifications](/tutorials/join-dal-baker/verify-rights).
+
+## Optional: Run an accuser
+
+The accuser is a daemon that monitors blocks and looks for problems, such as bakers who double-sign blocks or inject multiple attestations.
+If it finds a problem, it posts a denunciation operation, which penalizes the baker.
```
NicNomadic, line 137:
```suggestion
If it finds a problem, it posts a denunciation operation, which leads to penalizing the misbehaving baker.
```

docs/tutorials/join-dal-baker/run-baker.md
```diff
@@ -99,3 +130,34 @@ Therefore, staking more tez brings more rewards but does not reduce the attestat
...
+You don't have to run an accuser, but if you do, you can receive some of the penalties when the accuser denounces a baker.
```
NicNomadic, line 138:
```suggestion
You don't have to run an accuser, but if you do, you can receive as a reward part of the penalties of the denounced baker.
```

NicNomadic (approved):
LGTM.

docs/tutorials/join-dal-baker/run-baker.md
```diff
@@ -2,23 +2,54 @@
...
+1. Optional: Set up a remote signer to secure the keys that the baker uses as described in [Signer](https://tezos.gitlab.io/user/key-management.html#signer) in the Octez documentation.
 
 1. To run a baking daemon that connects to the DAL, start it as usual and pass the URL to your DAL node to it with the `--dal-node` argument:
 
    ```bash
    octez-baker-PsParisC run with local node "$HOME/.tezos-node" my_baker --liquidity-baking-toggle-vote pass --adaptive-issuance-vote on --dal-node http://127.0.0.1:10732
    ```
 
+   Note that the command for the baker depends on the protocol version.
+   This example uses the ParisC protocol, so the command starts with `octez-baker-PsParisC`.
```
NicNomadic, line 20: Is there a way to use a variable here for the protocol suffix that gets instantiated by Docusaurus at build time, to avoid changing the tutorial  at a protocol shift?

timothymcmackin: There are a few ways to do this -- I'll add a note as part of our discussion about a list of to-dos for each protocol upgrade.

PR #498, opened 2024-12-31
Update baker tutorial to use consensus keys

timothymcmackin:
It's better security practice to use a consensus key to sign consensus operations. Update the tutorial:
- Create two keys and explain why
- Use the consensus key to sign operations
- Include a section on changing the key

### DIFF
--- a/docs/tutorials/join-dal-baker/prepare-account.md
+++ b/docs/tutorials/join-dal-baker/prepare-account.md
@@ -1,14 +1,31 @@
 ---
-title: "Step 2: Set up a baker account"
+title: "Step 2: Set up baker accounts"
 authors: Tezos core developers, Tim McMackin
 last_update:
-  date: 19 August 2024
+  date: 31 December 2024
 ---
 
-The baker needs a user account that stakes tez.
-In this section, you use the Octez client to create an account, register it as a delegate, and stake tez with it.
+In this section you use the Octez client to set up two accounts for your baker:
 
-1. Create or import an account in the Octez client.
+- The baker key itself (also called the manager key) stakes tez and registers as a delegate
+- The consensus key is the key that the baker uses to sign attestations, which are generally referred to as _consensus operations_
+
+:::note
+
+Using a separate consensus key is not required but it is good security practice.
+Signing consensus operations incurs no fees, so you can set up a baker with a key that has no tez.
+You can use this key on a remote machine and keep your baker key with the tez in a more secure location to reduce risk to your funds.
+
+If the consensus key is compromised or lost, you can create a new key and switch the baker to it without changing how your tez is staked and delegated and without moving your delegators and stakers to a new account.
+In this way you can avoid backing up the consensus key or store it in a Key Management System (KMS) or Hardware Security Module (HSM) where no one has access to its private key.
+
+For more information about consensus keys, see [Consensus key](https://tezos.gitlab.io/user/key-management.html#consensus-key) in the Octez documentation.
+
+:::
+
+In this section, you use the Octez client to create these two accounts and set them up for baking.
+
+1. Create or import an account in the Octez client to be the baker or manager account.
 The simplest way to get an account is to use the Octez client to randomly generate an account.
 This command creates an account and associates it with the `my_baker` alias:
 
@@ -16,7 +33,7 @@ This command creates an account and associates it with the `my_baker` alias:
    octez-client gen keys my_baker
    ```
 
-   The address of the generated account can be obtained with the following command:
+   You can get the address of the generated account with this command:
 
    ```bash
    octez-client show address my_baker
@@ -47,10 +64,19 @@ This command creates an account and associates it with the `my_baker` alias:
 
    When the account receives its tez, it owns enough stake to bake but has still no consensus or DAL rights because it has not declared its intention to become a baker.
 
-1. Register your account as a delegate by running the following command:
+1. Set up a separate account to be the consensus key.
+This command creates an account and associates it with the `consensus_key` alias:
 
    ```bash
-   octez-client register key my_baker as delegate
+   octez-client gen keys consensus_key
+   ```
+
+   This account does not need any tez.
+
+1. Register the baker account as a delegate and set its consensus key by running the following command:
+
+   ```bash
+   octez-client register key my_baker as delegate with consensus key consensus_key
    ```
 
 1. Stake at least 6,000 tez, saving a small amount for transaction fees, by running this command:
@@ -59,7 +85,8 @@ This command creates an account and associates it with the `my_baker` alias:
    octez-client stake 6000 for my_baker
    ```
 
-Now the account has staked enough tez to earn the right to make attestations, including attestations that data is available on the DAL.
-However, it does not receive these rights until the baking daemon is running and a certain amount of time has passed.
+Now the baker account has staked enough tez to earn the right to make attestations, including attestations that data is available on the DAL.
+Its consensus key is authorized to sign consensus operations on its behalf.
+However, the baker account does not receive these rights until the baking daemon is running and a certain amount of time has passed.
 
 While you wait for attestation rights, continue to [Step 3: Run an Octez DAL node](/tutorials/join-dal-baker/run-dal-node).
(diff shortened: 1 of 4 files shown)


docs/tutorials/join-dal-baker/prepare-account.md
```diff
@@ -1,22 +1,39 @@
...
+If the consensus key is compromised or lost, you can create a new key and switch the baker to it without changing how your tez is staked and delegated and without moving your delegators and stakers to a new account.
+In this way you can avoid backing up the consensus key or store it in a Key Management System (KMS) or Hardware Security Module (HSM) where no one has access to its private key.
+
+For more information about consensus keys, see [Consensus key](https://tezos.gitlab.io/user/key-management.html#consensus-key) in the Octez documentation.
+
+:::
+
+In this section, you use the Octez client to create these two accounts and set them up for baking.
+
+1. Create or import an account in the Octez client to be the baker or manager account.
```
NicNomadic, line 28: just to be clear that baker account = manager account:
```suggestion
1. Create or import an account in the Octez client to be the baker account (sometimes called "manager" account).
```

docs/tutorials/join-dal-baker/prepare-account.md
```diff
@@ -59,7 +85,8 @@ This command creates an account and associates it with the `my_baker` alias:
    octez-client stake 6000 for my_baker
    ```
 
-Now the account has staked enough tez to earn the right to make attestations, including attestations that data is available on the DAL.
-However, it does not receive these rights until the baking daemon is running and a certain amount of time has passed.
+Now the baker account has staked enough tez to earn the right to make attestations, including attestations that data is available on the DAL.
+Its consensus key is authorized to sign consensus operations on its behalf.
+However, the baker account does not receive these rights until the baking daemon is running and a certain amount of time has passed.
```
NicNomadic, line 90: I think you receive rights even if the baker daemon is not running:
```suggestion
However, the baker account does not receive these rights until a certain amount of time has passed.
```

docs/tutorials/join-dal-baker/run-node.md
NicNomadic, line 1: Nothing to do with the consensus key, but in step "Load the snapshot" shouldn't you add a mention of an already filled context? E.g.:

If the import command raises an error of the form:

   ```
   octez-node: Error:
              Invalid data directory '.../.tezos-node': Please provide a clean directory by removing the following files: context, daily_logs, lock, store.
   ```

it may indicate that your node directory is already containing some context, so you should first remove the indicated files and directories.

NicNomadic, line 1: When one runs the tutorial not for the first time, the command `octez-client get balance for my_baker` may return less than 6Ktez spendable, but more than 6Ktez are already staked. Is there a command to get the already staked amount, without using a block explorer?

timothymcmackin: Added "If you get an error that says that the data directory is invalid, verify that the directory exists and is empty."

NicNomadic, line 1: In step "Start the node" or before it, shouldn't we tell, as you do for the DAL node, that the port where the nodes is listening for connections from other nodes (9732 by default) is accessible from the outside? If so, refer to <https://tezos.gitlab.io/user/node-configuration.html#listening-ports> for more info.

timothymcmackin: @NicNomadic I've done some refactoring here to account for people using accounts with existing stake; let me know what you think.

timothymcmackin: Added a note for the layer 1 node about making sure that the port is accessible.

NicNomadic: It always exists and non-empty, because it contains at least the config file created as told above. I think it's better to either:
- give the error sample (maybe simplified) and tell to follow the instructions
- or tell something like "If you get an error that says that the data directory is invalid, clean up the directory as indicated."

docs/tutorials/join-dal-baker/run-baker.md
NicNomadic, line 1: "You may need to install the curl program." comes too late here, it should be moved to file run-dal-node, where curl is first used.

docs/tutorials/join-dal-baker/run-node.md
timothymcmackin: > If you get an error that says that the data directory is invalid, clean up the directory as the error message indicates.

docs/tutorials/join-dal-baker/run-baker.md
timothymcmackin: Corrected.

docs/tutorials/join-dal-baker/prepare-account.md
```diff
@@ -1,65 +1,103 @@
...
-   octez-client get balance for my_baker
+   octez-client stake 6000 for my_baker
    ```
 
-   If the balance still shows 0, the local node may not be ready yet.
-   In this case you can temporarily use the public RPC endpoint.
+   If you are using a testnet and need tez to stake, you can get tez from the testnet faucet.
+   For example, if you are using Ghostnet, use the Ghostnet faucet linked from https://teztnets.com/ghostnet-about to send tez to the baker account.
+
+   If you send tez to the account from a faucet and the `get balance` command still shows 0, the local node may not be ready yet.
```
NicNomadic, line 69: I think that when the node is ready, you don't get 0, but just an error like "protocol not found" or something. I'd suggest removing this explanation (lines 69-74), because the previous page ensured that the node is bootstrapped.

docs/tutorials/join-dal-baker/prepare-account.md
```diff
@@ -1,65 +1,103 @@
...
+   If you need to stake more, pass the amount to the `stake` command, as in this example:
 
    ```bash
-   octez-client get balance for my_baker
+   octez-client stake 6000 for my_baker
    ```
 
-   If the balance still shows 0, the local node may not be ready yet.
-   In this case you can temporarily use the public RPC endpoint.
+   If you are using a testnet and need tez to stake, you can get tez from the testnet faucet.
```
NicNomadic, line 66: This mention of the faucet should go upper, in step "creating the baker account". Together with the caveat that it may take long.

docs/tutorials/join-dal-baker/run-node.md
NicNomadic: FIne with that command, thanks!

docs/tutorials/join-dal-baker/run-node.md
```diff
@@ -36,6 +36,9 @@ If you build from source, you can use the `latest-release` branch to work with G
 
 ## Running the layer 1 node
 
+1. Ensure that the port that the node runs on (by default, 8732) is accessible from outside its system.
```
NicNomadic, line 39: Not 8732, this one is for RPCs, and it's fine to be used only locally (by our baker and DAL node).
We're talking here about "the port where the node is listening for connections from peer nodes (9732 by default)"

docs/tutorials/join-dal-baker/run-node.md
NicNomadic: left a comment on the new version

docs/tutorials/join-dal-baker/prepare-account.md
timothymcmackin: Removed this section.

docs/tutorials/join-dal-baker/prepare-account.md
timothymcmackin: It makes more sense here because we've added the info about whether you may have already staked. This catches the case of a user who has staked some tez with another baker but now wants to stake enough to bake on their own. If we move the faucet info up they won't know how much to get from the faucet.

NicNomadic (commented):
I ran the tutorial up to Verifications (excluded), as I'm waiting for the baking rights. Will do these final checks on Monday.

NicNomadic:
> I ran the tutorial up to Verifications (excluded), as I'm waiting for the baking rights. Will do these final checks on Monday.

Having got rights yesterday, I could run the tutorial to completion, and I can confirm everything worked as expected.

=> LGTM

NicNomadic (approved):

PR #499, opened 2024-12-31
Image test updates

timothymcmackin:
1. Expand the test to check for images in html `<img` tags
2. Expand the test to get a list of image files and check to see if they are used.
3. Remove the images that the test flagged as unused.

To test item 1, delete the file `static/img/using/staking-current-balance.png` and run the test with `npm run  test`. It should detect that this file is missing.

To test item 2, put an image file in `static/` and the test should show that it is present but not used in any file.


beatalipska (approved):
💯

PR #511, opened 2025-01-15
Provide example systemd file names and commands

timothymcmackin:
- Use protocol-specific file names and service names for systemd services because during upgrade you'll have two bakers and two accusers running
- Provide example file names and example commands for using the service files with systemd

### DIFF
--- a/docs/tutorials/join-dal-baker/run-baker.md
+++ b/docs/tutorials/join-dal-baker/run-baker.md
@@ -57,7 +57,7 @@ You can also refer to [Run a persistent baking node](https://opentezos.com/node-
    SyslogIdentifier=%n
    ```
 
-   If you name this service file `/etc/systemd/system/octez-baker-PsQuebec`, you can start it by running these commands:
+   If you name this service file `/etc/systemd/system/octez-baker-PsQuebec.service`, you can start it by running these commands:
 
    ```bash
    sudo systemctl daemon-reload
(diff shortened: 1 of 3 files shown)


timothymcmackin:
@germanD may be a good choice of reviewer for this item.

docs/tutorials/join-dal-baker/run-baker.md
```diff
@@ -57,6 +57,19 @@ You can also refer to [Run a persistent baking node](https://opentezos.com/node-
    SyslogIdentifier=%n
    ```
 
+   If you name this service file `/etc/systemd/system/octez-baker-PsQuebec`, you can start it by running these commands:
```
windushka, line 60: The instruction mentions naming the service file /etc/systemd/system/octez-baker-PsQuebec. Shouldn't it include the .service extension for clarity?
```suggestion
   If you name this service file `/etc/systemd/system/octez-baker-PsQuebec.service`, you can start it by running these commands:
```

docs/tutorials/join-dal-baker/run-node.md
```diff
@@ -129,6 +129,35 @@ You can also refer to [Run a persistent baking node](https://opentezos.com/node-
    SyslogIdentifier=%n
    ```
 
+   If you name this service file `/etc/systemd/system/octez-node`, you can start it by running these commands:
```
windushka, line 132: Same here?
```suggestion
   If you name this service file `/etc/systemd/system/octez-node.service`, you can start it by running these commands:
```

windushka (changes requested):

windushka (approved):

PR #519, opened 2025-01-17
Must use absolute links to files not processed by Docusaurus

timothymcmackin:
Per the warning at the bottom of [this page](https://docusaurus.io/docs/markdown-features/links), you can't link from a file that Docusaurus processes to a file that it does not process with a relative or root-relative link. For this reason, when you go to https://docs.tezos.com/textfiles and click on a file name, the site will show a 404 even though the path is correct. Setting the links to absolute https links fixes this.


NicNomadic (approved):
Well spotted, LGTM

PR #521, opened 2025-01-17
[wip] Set dependncy versions

timothymcmackin:
https://github.com/trilitech/tezos-developer-docs/pull/503 introduced a test for dependency versions. This PR adds the metadata on the latest versions of dependencies that procedures were tested with.


docs/dApps/wallets.md
```diff
@@ -3,6 +3,8 @@ title: Connecting to wallets
 authors: "Claude Barde, Tim McMackin"
 last_update:
   date: 6 November 2023
+dependencies:
+  taquito: 21.0.1
```
NicNomadic, line 7: do we need minor sub-version numbers? I mean, wouldn't is enough to say 21.0? (questions is for all the other tools, too)

NicNomadic: Or maybe it could make sense to leave precise numbers like this, but maybe (optionally?) ignore the last number when comparing with the current version?

src/scripts/dependencies.json
```diff
@@ -1,12 +1,16 @@
 {
   "comment": "These are the current versions of tools used in the check_dependencies script.",
   "currentVersions": {
-    "archetype": "1.0.26",
-    "octez-client": "20.1",
+    "completium-cli": "1.0.26",
+    "ligo": "1.9.2",
+    "octez": "21.2",
```
NicNomadic, line 6: It can be annoying to update this at each Octez minor release. Wouldn't be more practical to have here only the major version (21 for now) and consider any tutorial tested on 21.* as up to date?

docs/dApps/wallets.md
timothymcmackin: How much detail to go into is up to us. On the side of the individual pages it's easy enough to put in the complete major-minor-fixpack version because you'll have it as you make the updates. In the list of current versions, sure, maybe we could limit it to the major versions. It depends on how we want to use these numbers and if we want to run tests based on them instead of just printing the info like we do now.

NicNomadic: Agreed

src/scripts/dependencies.json
NicNomadic: OK, we said that the check would ignore the minor version for Octez, so this remains to be implemented, but here the version number is correct.

NicNomadic (approved):
LGTM modulo a followup to ignore minor version numbers for Octez (and perhaps tje last number in all the other dependencies).

src/scripts/dependencies.json
timothymcmackin: I think a better way to do this is to put all the metadata in the files and use parameters to determine what level of dependency we care about: https://github.com/trilitech/tezos-developer-docs/pull/534

Issue #527, opened 2025-01-23
[BUG] Delegate comes before stake


ajinkyaraj-23:
**Describe the bug**
In step 2 of the setup a baker tutorial, stake command is written before register as delegate.
But in reality the baker has to register himself as a delegate( or baker). Then only he can execute stake command. 

**To Reproduce**
1. create a new key with some balance
2. Try to stake tez, before registering as delegate , it will throw error. 

**Expected behavior**
Make this sequence clear to the user. 

**Screenshots**

timothymcmackin:
Fixed.

PR #532, opened 2025-01-24
fix order of baker account setup (issue #527)

NicNomadic:
Fix issue #527 by reordering the operations for setting up a baker account, in the baking tutorial page <https://docs.tezos.com/tutorials/join-dal-baker/prepare-account>.

### DIFF
--- a/docs/tutorials/join-dal-baker/prepare-account.md
+++ b/docs/tutorials/join-dal-baker/prepare-account.md
@@ -50,12 +50,6 @@ This command creates an account and associates it with the `my_baker` alias:
 
 1. Ensure your account has at least 6,000 tez to stake, plus a small liquid amount for transaction fees.
 
-   You can check the full balance of the account (staked + non-staked) with this command:
-
-   ```bash
-   octez-client get full balance for my_baker
-   ```
-
    If you are using a testnet and need more tez, you can get tez from the testnet faucet.
    For example, if you are using Ghostnet, use the Ghostnet faucet linked from https://teztnets.com/ghostnet-about to send tez to the baker account.
 
@@ -85,16 +79,22 @@ This command creates an account and associates it with the `consensus_key` alias
 1. Stake at least 6,000 tez with the account, saving a small liquid amount for transaction fees.
 Staked tez is temporarily frozen and cannot be spent, so you need some unstaked tez to pay transaction fees.
 
+   Pass the amount to the `stake` command, as in this example:
+
+   ```bash
+   octez-client stake 6000 for my_baker
+   ```
+
    You can check how much you have staked by running this command:
 
    ```bash
    octez-client get staked balance for my_baker
    ```
 
-   If you need to stake more, pass the amount to the `stake` command, as in this example:
+   You can also check the full balance of the account (staked + non-staked) with this command:
 
    ```bash
-   octez-client stake 6000 for my_baker
+   octez-client get full balance for my_baker
    ```
 
 Now the baker account has staked enough tez to earn the right to make attestations, including attestations that data is available on the DAL.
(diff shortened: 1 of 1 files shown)


docs/tutorials/join-dal-baker/prepare-account.md
```diff
@@ -89,6 +82,21 @@ This command creates an account and associates it with the `consensus_key` alias
 
    If you are not using a consensus key, omit the argument `with consensus key consensus_key`.
 
+1. Stake at least 6,000 tez with the account, saving a small liquid amount for transaction fees.
```
ajinkyaraj-23, line 85:
```suggestion
2. Stake at least 6,000 tez with the account, saving a small liquid amount for transaction fees.
```

docs/tutorials/join-dal-baker/prepare-account.md
```diff
@@ -89,6 +82,21 @@ This command creates an account and associates it with the `consensus_key` alias
...
+Staked tez is temporarily frozen and cannot be spent, so you need some unstaked tez to pay transaction fees.
+
+   You can check how much you have staked by running this command:
+
+   ```bash
+   octez-client get staked balance for my_baker
```
ajinkyaraj-23, line 91: The user has not staked until now, so this is counterintuitive. 
put the stake 6000 command first and then 
To check your staked amount use this command.

docs/tutorials/join-dal-baker/prepare-account.md
timothymcmackin: We can make all the numbered steps as 1 because Markdown will put the numbers in the correct order in the output.

ajinkyaraj-23:
He would have got that info from full balance command

docs/tutorials/join-dal-baker/prepare-account.md
NicNomadic: Added a note for people running the tutorial for a second time.

NicNomadic:
> He would have got that info from full balance command

The full balance gives you the sum of staked + non-staked, so users need to know how to get the sum and each component, right?

ajinkyaraj-23:
Frozen balance gives division as well.

ajinkyaraj-23:
Also it is not possible for users to have staked balance without first
delegating themselves as baker

NicNomadic:
> Also it is not possible for users to have staked balance without first delegating themselves as baker

Again, this case happens when users don't run the tutorial for the first time.

NicNomadic:
> Frozen balance gives division as well.

Sorry, I don't understand what you mean.

ajinkyaraj-23:
> > Also it is not possible for users to have staked balance without first delegating themselves as baker
> 
> Again, this case happens when users don't run the tutorial for the first time.

If they are going through this tutorial again, they would know how to check for balance. I am of opinion that staking would come before checking the staked balance.

ajinkyaraj-23:
> > Frozen balance gives division as well.
> 
> Sorry, I don't understand what you mean.

get full balance command gives total balance, frozen balance and unfrozen balance.

NicNomadic:
> > > Also it is not possible for users to have staked balance without first delegating themselves as baker
> > 
> > 
> > Again, this case happens when users don't run the tutorial for the first time.
> 
> If they are going through this tutorial again, they would know how to check for balance. I am of opinion that staking would come before checking the staked balance.

You're right. I changed the order to stake before checking the staked and full balances.

ajinkyaraj-23 (approved):
LGTM

timothymcmackin (approved):

docs/tutorials/join-dal-baker/prepare-account.md
NicNomadic: no more applicable

PR #534, opened 2025-01-28
Allow filtering by version level

timothymcmackin:
The current dependency script alerts on any version difference up to the fixpack level. This change allows you to pass `--major` to show only places where dependencies are out of date by the major version or `--minor` for minor version.

For example, `npm run check-dependencies -- --major octez` prints this:

```
docs/tutorials/smart-rollup.md
Outdated: octez 19 < 21.2

Reviewed 154 files.
```

But `npm run check-dependencies -- --minor octez` prints this:

```
docs/developing/octez-client/accounts.md
Outdated: octez 21.1 < 21.2

docs/developing/octez-client/installing.md
Outdated: octez 21.1 < 21.2

docs/developing/octez-client/transactions.md
Outdated: octez 21.1 < 21.2

docs/tutorials/smart-rollup.md
Outdated: octez 19 < 21.2

Reviewed 154 files.
```

### DIFF
--- a/src/scripts/check_dependencies.mjs
+++ b/src/scripts/check_dependencies.mjs
@@ -28,7 +28,7 @@ import semver from 'semver';
 import minimist from 'minimist';
 
 const argv = minimist(process.argv.slice(2), {
-  boolean: ['major', 'minor', 'fixpack'],
+  boolean: ['major', 'minor'],
 });
 const params = argv['_'];
 
--- a/src/scripts/check_dependencies.mjs
+++ b/src/scripts/check_dependencies.mjs
@@ -27,32 +27,31 @@ import matter from 'gray-matter';
 import semver from 'semver';
 import minimist from 'minimist';
 
-const argv = minimist(process.argv.slice(2), {
-  boolean: ['major', 'minor'],
-});
-const params = argv['_'];
-
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const baseFolder = path.resolve(__dirname, '../..');
 const docsFolder = path.resolve(baseFolder, 'docs');
-
-const checkMajor = argv.major;
-const checkMinor = argv.minor;
-if (checkMajor && checkMinor) {
-  console.error('Include either --major or --minor, not both');
-  process.exit(1);
-}
 
 // Import config file
 // Not sure why I can't import a JSON file directly in MJS
 const dependencyConfig = fs.readFileSync(path.resolve(__dirname, 'dependencies.json'), 'utf8');
 const { currentVersions } = JSON.parse(dependencyConfig);
 
-// Verify the passed dependencies parameters
-const unrecognizedParams = params.filter((oneParam) => !currentVersions[oneParam]);
-if (unrecognizedParams.length > 0) {
-  console.error('Unrecognized tool names in parameters:');
-  console.log(unrecognizedParams.join('\n'));
+const argv = minimist(process.argv.slice(2), {
+  boolean: ['major', 'minor'],
+  unknown: (unknownArg) => {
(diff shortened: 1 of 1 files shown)


src/scripts/check_dependencies.mjs
```diff
@@ -8,6 +8,15 @@ dependencies:
...
 ---
 
+List the current versions of the tools in the config file at ./dependencies.json and this script will print the pages that have been tested only on older versions of the tools.
+
+Arguments:
+
+You can pass individual tools to check, as in npm run check-dependencies -- smartpy taquito and it will check only those dependencies.
+
+By default, this script checks for differences down to the fixpack level.
+You can pass --major or --minor to ignore differences up to the specified level.
```
NicNomadic, line 18: I see that you can also pass `--fixpack`, right?

src/scripts/check_dependencies.mjs
```diff
@@ -18,18 +27,21 @@ import matter from 'gray-matter';
...
 const docsFolder = path.resolve(baseFolder, 'docs');
 
-// List the current versions of the tools in the config file and this script will print the pages that have been tested only on older versions of the tools
-
-// You can pass individual tools to check, as in
-// npm run check-dependencies smartpy taquito
-// and it will check only those dependencies
+const checkMajor = argv.major;
+const checkMinor = argv.minor;
+if (checkMajor && checkMinor) {
```
NicNomadic, line 41: Shouldn't you also check overlap with `--fixpack`?

Note that you  could remove the need for all this kind of checks if the UI were "--level=major/minor/fixpack"

NicNomadic (commented):
Nice result!

As a side node, I thought the UI was more robust in the past, didn't t check for unknown dependencies? Now it doesn't seem to warn in case of misspelled options and/or arguments:

    tezos-developer-docs % npm run check-dependencies -- --blabla zvelt

lists all the outdated dependecies, without complaining about anything.

src/scripts/check_dependencies.mjs
timothymcmackin: I meant to remove the `--fixpack` argument because that is the default. I'll take it out of the minimist config now.

timothymcmackin:
> Nice result!
> 
> As a side node, I thought the UI was more robust in the past, didn't t check for unknown dependencies? Now it doesn't seem to warn in case of misspelled options and/or arguments:
> 
> ```
> tezos-developer-docs % npm run check-dependencies -- --blabla zvelt
> ```
> 
> lists all the outdated dependecies, without complaining about anything.

In this case you are passing an argument `--blabla` with the value `zvelt` and it's swallowing those. I'll add an error to catch unrecognized arguments. It still checks for unknown dependencies if you run something like `npm run check-dependencies -- blabla zvelt`.

src/scripts/check_dependencies.mjs
NicNomadic: Oh I see! Besides, it was written just above, never mind, sorry.

src/scripts/check_dependencies.mjs
NicNomadic: will be solved by your check of unexpected arguments (or "option names")

timothymcmackin: `--fixpack` would be the default, but we could include it for completeness. I think these boolean switches are easier than `--arg=value` because then I don't have to remember whether it's `--arg=value` or `--arg value`.

NicNomadic: I agree. And my initial concern is solved by your new check, thanks.

NicNomadic (approved):
LGTM

PR #537, opened 2025-01-29
More info about recovering bond

timothymcmackin:
Per feedback, provide some more info about recovering bonds.

### DIFF
--- a/docs/architecture/smart-rollups.md
+++ b/docs/architecture/smart-rollups.md
@@ -2,7 +2,7 @@
 title: Smart Rollups
 authors: 'Nomadic Labs, TriliTech, Tim McMackin'
 last_update:
-  date: 17 January 2025
+  date: 29 January 2025
 ---
 
 Smart Rollups play a crucial part in providing high scalability on Tezos.
@@ -87,8 +87,10 @@ When a user runs a node that posts commitments, the protocol automatically locks
 If the node posts a commitment that is refuted, they lose their bond, as described in [Refutation periods](#refutation-periods).
 
 Because nodes have the length of the refutation to challenge another node's commitment, the bond stays locked until the end of the refutation period for the last commitment that the node posted.
-After that refutation period ends, the node operator can recover the bond, which unlocks their tez.
+After that refutation period ends, the node operator can recover the bond with the `octez-client recover bond` command, which unlocks their tez.
 To simplify the process, node operators can switch to bailout mode, which does not post commitments but continues to defend previously made commitments until the last refutation period ends.
+Then it runs the `octez-client recover bond` command automatically.
+For an example of how to use bailout mode to recover your bond, see [Stopping the Smart Rollup node](https://docs.etherlink.com/network/smart-rollup-nodes/#stopping-the-smart-rollup-node) in the Etherlink documentation.
 
 ### Reveal data channel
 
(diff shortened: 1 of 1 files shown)


docs/architecture/smart-rollups.md
```diff
@@ -87,8 +87,10 @@ When a user runs a node that posts commitments, the protocol automatically locks
 If the node posts a commitment that is refuted, they lose their bond, as described in [Refutation periods](#refutation-periods).
 
 Because nodes have the length of the refutation to challenge another node's commitment, the bond stays locked until the end of the refutation period for the last commitment that the node posted.
-After that refutation period ends, the node operator can recover the bond, which unlocks their tez.
+After that refutation period ends, the node operator can recover the bond with the `octez-client recover bond` command, which unlocks their tez.
 To simplify the process, node operators can switch to bailout mode, which does not post commitments but continues to defend previously made commitments until the last refutation period ends.
+Then it runs the `octez-client recover bond` command automatically.
```
NicNomadic, line 92: Multiple remarks:
- "which unlock their tez" gives the impression that they are unlocked immediately, while it's just a request 
- "it runs the command automatically" seems to contradict the fact that you already executed the command
- in this page, it's not clear that there are two possible ways to do it: bailout mode or recover command; this seems pretty clear in the other page.
- "After the refutation period ends" is contradicted by " until the last refutation period ends"

So I see two possibilities:
1. remove the operational details here and just refer to the Etherlink doc
2. keep some details here, reformulating them more clearly, e.g.:
 ```suggestion
The node operator can recover the bond in different ways:
- using the `octez-client recover bond` command, which unlocks their tez after the refutation period ends.
- alternatively, by switching the rollup node to bailout mode, which does not post commitments but continues to defend previously made commitments until the last refutation period ends, then runs the `octez-client recover bond` command automatically.
```

File: docs/architecture/smart-rollups.md [hunk 2]
```diff
@@ -88,8 +88,10 @@ If the node posts a commitment that is refuted, they lose their bond, as describ
 
 Because nodes have the length of the refutation to challenge another node's commitment, the bond stays locked until the end of the refutation period for the last commitment that the node posted.
 The node operator can recover the bond in different ways:
-- using the `octez-client recover bond` command, which unlocks their tez after the refutation period ends.
-- alternatively, by switching the rollup node to bailout mode, which does not post commitments but continues to defend previously made commitments until the last refutation period ends, then runs the `octez-client recover bond` command automatically.
+
+- Using the `octez-client recover bond` command, which unlocks their tez after the refutation period ends.
```
NicNomadic, line 92: In fact, does this way really work? AFAIU from your explanations, if you do this command:
- ... and leave the SR node running, it will continue to post commitments so the refutation period may never end
- ... and stop the SR node, you risk losing your bond in a refutation game
So the only safe option is to switch the SR node to bailout mode, which is the way #2.

Is there something I don't understand? In fact, what is the use case of running the command manually?

NicNomadic: By looking to the mode descriptions at <https://tezos.gitlab.io/shell/smart_rollup_node.html#deploying-a-rollup-node>, it seems to me that this command can work only if you switch the SR node to accuser mode. Is this right? (If so, we should tell it)

timothymcmackin: Yes, we either need to remove the manual option of running the command or make it clear that you can do this only after the node has stopped making commitments.

I don't see anywhere at that link where it says that you can run the command only if the node is in accuser mode. That doesn't make sense to me; even if I stop my node I should be able to recover my bond.

NicNomadic: It doesn't talk about bond recovery, but by looking at the accuser mode it seems it's what you need for doing it. AFAIU, it's essentially like the bailout mode but without the automatic call to the bond recovery command at the end, so running it + manual bond recovery seems to work, isn't it?

timothymcmackin: Reworked it to use accuser or bailout mode.

NicNomadic: Perfect now, crystal clear!

NicNomadic (approved):

PR #538, opened 2025-01-31
Allow paths or specific files

timothymcmackin:
For dependency checks we should be able to pass individual paths or files. This PR allows you to pass any combination, as in these examples:

```
npm run check-dependencies -- --major -d=smartpy,taquito docs/developing/octez-client/accounts.md docs/tutorials

npm run check-dependencies -- --major -d=smartpy,taquito docs/*
```

The shell expands the `*` in the anonymous arguments before it gets to the script but I think I've got it working despite that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ When you run `npm run check-dependencies`, a script checks for files that need t
 
 - You can check specific tools by passing them to the script, as in `npm run check-dependencies -- --dependencies=smartpy,taquito`.
 - By default, the script checks for differences down to the fixpack level, but you can pass `--major` or `--minor` to check for differences at those levels.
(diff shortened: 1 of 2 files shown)


NicNomadic:
Is this extension necessary? The whole point for passing arguments expanded by the shell (that you already implemented) is to count on the support of orthogonal tools to do it for you, according to the general philosophy of Unix. For instance, if we want to scan all the MD files in "docs/" we can:

> find docs -name "*.md" | xargs npm run check-dependencies -- --major

timothymcmackin:
This makes the script easier to use and doesn't prevent you from piping data from another program like that.

NicNomadic (approved):
> This makes the script easier to use and doesn't prevent you from piping data from another program like that.

OK then.

PR #543, opened 2025-02-07
Backup and restore instructions for nonce files

timothymcmackin:
Need to backup nonce files if you move the baker to another machine.

### DIFF
--- a/docs/tutorials/join-dal-baker/run-baker.md
+++ b/docs/tutorials/join-dal-baker/run-baker.md
@@ -129,7 +129,7 @@ You can upgrade accusers with a similar process.
 The Octez baking daemon stores persistent operational data in the Octez client's data directory, notably consensus high-water marks and [random seed nonces](https://tezos.gitlab.io/active/randomness_generation.html).
 If you want to back up the baker or move it to another machine and restore it, you must copy the nonce file or files from the Octez client's data directory to the equivalent directory on the new machine.
 These nonce files are named `net<NETWORK_ID>_stateful_nonces` and `net<NETWORK_ID>_nonces`, where `<NETWORK_ID>` is the ID of the network.
-All systems have the `net<NETWORK_ID>_stateful_nonces` file but only legacy bakers have the `net<NETWORK_ID>_nonces` file.
+All systems have the `net<NETWORK_ID>_stateful_nonces` file but only legacy baking deployments have the `net<NETWORK_ID>_nonces` file.
 
 ## Calculating the delay for attestation rights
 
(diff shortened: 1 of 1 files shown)


docs/tutorials/join-dal-baker/run-baker.md
```diff
@@ -124,6 +124,13 @@ However, make sure not to run the baker twice **for the same protocol version an
 
 You can upgrade accusers with a similar process.
 
+## Backing up and restoring the baker
+
+The Octez baking daemon stores no persistent data of its own, but it depends on files in the Octez client's data directory.
+If you want to back up the baker or move it to another machine and restore it, you must copy the nonce file or files from the Octez client's data directory to the equivalent directory on the new machine.
+These nonce files are named `net<NETWORK_ID>_stateful_nonces` and `net<NETWORK_ID>_nonces`, where `<NETWORK_ID>` is the ID of the network.
+All systems have the `net<NETWORK_ID>_stateful_nonces` file but only legacy bakers have the `net<NETWORK_ID>_nonces` file.
```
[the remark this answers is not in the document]
germanD: This is about how long your `client dir` has been in use, and whether it was set up before stateful nonces became the default.

docs/tutorials/join-dal-baker/run-baker.md
```diff
@@ -124,6 +124,13 @@ However, make sure not to run the baker twice **for the same protocol version an
 
 You can upgrade accusers with a similar process.
 
+## Backing up and restoring the baker
+
+The Octez baking daemon stores no persistent data of its own, but it depends on files in the Octez client's data directory.
+If you want to back up the baker or move it to another machine and restore it, you must copy the nonce file or files from the Octez client's data directory to the equivalent directory on the new machine.
+These nonce files are named `net<NETWORK_ID>_stateful_nonces` and `net<NETWORK_ID>_nonces`, where `<NETWORK_ID>` is the ID of the network.
```
germanD, line 131: Can we add an example for mainnet/ghostnet?

germanD (commented):
LGTM. 

As a follow up, we should check with the L1 bakewr crew if both files actually need backing up. My understanding is that with current versions of Octez keeping the _stateful_ one should suffice, and the other one can be safely removed. 

But I would rather be safe than sorry.

docs/tutorials/join-dal-baker/run-baker.md
```diff
@@ -124,6 +124,13 @@ However, make sure not to run the baker twice **for the same protocol version an
 
 You can upgrade accusers with a similar process.
 
+## Backing up and restoring the baker
+
+The Octez baking daemon stores persistent operational data in the Octez client's data directory, notably consensus high-water marks and [random seed nonces](https://tezos.gitlab.io/active/randomness_generation.html).
+If you want to back up the baker or move it to another machine and restore it, you must copy the nonce file or files from the Octez client's data directory to the equivalent directory on the new machine.
+These nonce files are named `net<NETWORK_ID>_stateful_nonces` and `net<NETWORK_ID>_nonces`, where `<NETWORK_ID>` is the ID of the network, such as `netXdQprcVkpaWU_stateful_nonces` for Mainnet or `netXnHfVqm9ie_stateful_nonces` for Ghostnet.
+All systems have the `net<NETWORK_ID>_stateful_nonces` file but only legacy baking deployments have the `net<NETWORK_ID>_nonces` file.
```
germanD, line 132:
```suggestion
All systems have the `net<NETWORK_ID>_stateful_nonces` file but only legacy baking deployments have the `net<NETWORK_ID>_nonces` file.
```
systems -> deployments. 

We should expand  legacy to mean, that were first deployed with Octez versions earlier/older than v20.0~rc1

germanD (approved):

PR #544, opened 2025-02-12
[baking-with-ledger] recommend to disable the `Battery saver`

spalmer25:
Some users have experienced problems while baking with their Ledger. It would appear that this is because the Ledger's battery saver option was activated and the Ledger shut itself down. This PR adds a section advocating the deactivation of this option.


ajinkyaraj-23 (approved):
LGTM

ajinkyaraj-23:
Hi @timothymcmackin Could you approve the deployment and then merge this in the docs.

timothymcmackin (approved):

